PR #9, opened 2018-02-07
Initial travis cfg for etherslurp

Martin2112:
It's meant to be lighweight. I haven't brought over all the baggage from the main repo.

And there appears to be no tests? Oh dear etc.


AlCutter:
Super neat - thanks!

AlCutter:
It was only meant to be a quick demo, I think @benlaurie might be working on something a little more refined :)

Martin2112:
At least this should check that it builds ok.

PR #18, opened 2018-04-25
Extract from log

benlaurie:
And add a Makefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,9 @@ coverage.txt
 
 # Terraform related files
 *.tfstate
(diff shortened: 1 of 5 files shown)


benlaurie:
Wrong repo!

Martin2112:
Are you sure it's the wrong repo? Looked like a trillian example to me!

Martin

benlaurie:
Yes, it is - but still a WIP. You can look at https://github.com/benlaurie/trillian-examples/pulls for the current status (each PR builds on the previous one, at least so far).

Martin2112:
Ah, OK. Carry on!

Martin

PR #57, opened 2019-04-08
Bitrot: delete removed LogRoot fields

AlCutter:


Martin2112:
Sorry I did go through the examples repo when I took these fields away. Must have missed one.

PR #72, opened 2019-09-13
Made examples build with go v1.13 and tidied modules

### LINKED ISSUE #71: Travis relies on MODPROXY=direct
The travis config uses MODPROXY=direct in order to work around cruftiness with 3rd party deps.
Once the deps are all fixed up then we should remove this workaround. 

Related to https://github.com/google/trillian/issues/1810

mhutchinson:
* Upgraded to latest version of certificate-transparency-go to avoid last known bad dependency.
* go mod tidy
* Use latest version of golangci to be consistent with other projects
* Updated Travis so it uses module proxy again

Fixes #71


mhutchinson:
This fixes #71 if it builds

PR #93, opened 2020-06-29
Fix govet issues

AlCutter:


Martin2112:
Just FYI I noticed the gossip example uses `FixedBackendResolver`. I've deleted this from the c-t-go repo so it will break when we bump the version.

I can apply the same fix as in the other repo to remove it. I'll let you merge this first though.

PR #94, opened 2020-06-30
Remove use of FixedBackendResolver, which has been deleted.

Martin2112:
The replacement should be functionality identical.

Also removed the deprecated way of setting up roundrobin balancing. This might need more work to be fully up to date with gRPC API changes but this change should avoid problems in using a newer version of gRPC.

### DIFF
--- a/gossip/hub/hub_server/main.go
+++ b/gossip/hub/hub_server/main.go
@@ -90,7 +90,7 @@ func main() {
 	dialOpts := []grpc.DialOption{grpc.WithInsecure()}
 	if strings.Contains(*rpcBackend, ",") {
 		// This should probably not be used in production. Either use etcd or a gRPC
-		// load balancer. 
+		// load balancer.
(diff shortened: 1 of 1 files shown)


gossip/hub/hub_server/main.go
```diff
@@ -89,10 +89,18 @@ func main() {
 
 	dialOpts := []grpc.DialOption{grpc.WithInsecure()}
 	if strings.Contains(*rpcBackend, ",") {
-		glog.Infof("Using FixedBackendResolver")
-		// Use a fixed endpoint resolution that just returns the addresses configured on the command line.
-		res := ctfeutil.FixedBackendResolver{}
-		dialOpts = append(dialOpts, grpc.WithBalancer(grpc.RoundRobin(res))) // nolint: megacheck
+		// This should probably not be used in production. Either use etcd or a gRPC
```
AlCutter, line 92: Could potentially just delete this support for passing in multiple backends?
It's only meant to be an example.

Martin2112: True, it probably wasn't being used. I've removed it.

PR #96, opened 2020-06-30
Try bumping trillian to 1.3.9.

Martin2112:
+ go mod tidy.


gdbelvin:
this repo has a dependency on github.com/google/certificate-transparency-go, which uses an old version of etcd
First thing to do is to update ct-go to the latest version of trillian

PR #102, opened 2020-07-08
Move Tritter into trillian-examples

AlCutter:


Martin2112:
Think this is moot now.

PR #103, opened 2020-07-08
Bump Trillian dep to v1.3.10

AlCutter:


Martin2112:
Looks like this needs the same gomock / gRPC proto fix we did for the other one.

Martin2112:
It's getting inconsistent etcd versions somehow.

```
go: downloading go.etcd.io/etcd v3.3.18+incompatible
go: finding github.com/coreos/etcd v3.3.22+incompatible
```

I don't see any direct references to the coreos version in examples.

Martin2112:
I think it might be because gossip pulls in monologue, which then pulls in c-t-go at an earlier version than trillian-examples does.

Perhaps we can fix this by making the gossip code a separate module and leaving it at the old version of things for now.

Or we could just disable the gossip test for the moment. We don't run this in production anywhere.

Martin2112:
A third option is to pull in the bits of monologue the test uses and break the dependency on monologue. However, I tested this and it won't work until there is a CT release that includes the changes that break the circular dependency. This is all less than ideal.

Martin2112:
Resolved by #104.

PR #106, opened 2020-07-10
Add the 'tritter' example to our examples repo

Martin2112:
+ Regenerate the protos with the same protoc and protoc-gen-go versions as travis uses.


Martin2112:
Hmm travis seems to be stuck.

Martin2112:
Based on a copy of similar PR by al to investigate build error

PR #136, opened 2020-11-16
Generate a verifiable map from sumdb auditor DB

mhutchinson:
This generates a map where the keys are the module+version and a component to indicate whether it is the repo hash or the go.mod hash, and the value is the hash value string as found in the go.sum file.

The verifier can (inefficiently) check that all values in a go.sum file are committed to in the verifiable map.


experimental/batchmap/sumdb/build/map.go
```diff
@@ -0,0 +1,157 @@
...
+
+	output := filepath.Clean(*output)
+	if output == "" {
+		glog.Exitf("No output provided")
+	}
+
+	// Create the directory if it doesn't exist
+	if _, err := os.Stat(output); os.IsNotExist(err) {
+		if err = os.Mkdir(output, 0700); err != nil {
+			glog.Fatalf("couldn't find or create directory %s, %v", output, err)
```
AlCutter, line 84: Exitf?

AlCutter, line 84: Also could use %q for dir (and error?)

AlCutter: some more below too

experimental/batchmap/sumdb/build/map.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/golang/protobuf/proto"
+
+	"github.com/google/trillian/experimental/batchmap"
+	"github.com/google/trillian/experimental/batchmap/tilepb"
+	"github.com/google/trillian/merkle/coniks"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+const hash = crypto.SHA512_256
```
AlCutter, line 43: Could this be:
```golang
var hash = sha512.Sum512_256
```

and below just use hash(...) rather than instantiating a new Hasher?

(same in `verify.go`)

experimental/batchmap/sumdb/verify/verify.go
```diff
@@ -0,0 +1,198 @@
...
+	mapDir       = flag.String("map_dir", "", "Directory containing map tiles.")
+	treeID       = flag.Int64("tree_id", 12345, "The ID of the tree. Used as a salt in hashing.")
+	prefixStrata = flag.Int("prefix_strata", 2, "The number of strata of 8-bit strata before the final strata.")
+)
+
+func main() {
+	flag.Parse()
+
+	if *mapDir == "" {
+		glog.Fatal("No map_dir provided")
```
AlCutter, line 50: Exitf throughout?

experimental/batchmap/sumdb/verify/verify.go
```diff
@@ -0,0 +1,198 @@
...
+	// This is optimized for clarity of demonstrating inclusion, not for performance.
+	// This code leads to a lot of duplicate I/O and tile calculation, particularly
+	// for the higher tiles in the map (e.g. the root tile will be read and verified
+	// for every single line in the go.sum file).
+	for scanner.Scan() {
+		line := scanner.Text()
+		split := strings.LastIndex(line, " ")
+		key := line[:split]
+		expectedString := line[split+1:]
+		glog.V(1).Infof("checking key '%s' value '%s'", key, expectedString)
```
AlCutter, line 77: can use '%q' in a bunch of places in this file rather than `'%s'`

AlCutter (approved):
Just a few minor nits inline.

experimental/batchmap/sumdb/build/map.go
mhutchinson: We still need this `crypto.Hash` to pass into `batchmap.Create`. There's also the problem that in Beam having variables around requires them to be serializable to be run inside a ParDo function. Constants are easier.

What I may do in the future is come back to this and optimize it for performance, but right now it's designed to be as clear as possible and have as few pointy ends to poke your eyes out around the Beam machinery as possible.

mhutchinson:
Cheers. Comments addressed and I snuck in a README.

PR #143, opened 2020-11-19
Split client into submit and readonly clients

AlCutter:


mhutchinson (approved):

binary_transparency/firmware/internal/client/client.go
```diff
@@ -32,16 +32,21 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// Client is an HTTP client for the FT personality.
+// ReadonlyClient is an HTTP client for the FT personality.
 //
 // TODO(al): split this into Client and SubmitClient.
-type Client struct {
+type ReadonlyClient struct {
```
mhutchinson, line 38: Only minor comment is whether ReadonlyClient or ReadOnlyClient. But whatever :-)

AlCutter: I think `readonly` is one word, so I'm gonna run with this for now :)

PR #144, opened 2020-11-19
Factor out devices bits from cmds

AlCutter:


mhutchinson (approved):

binary_transparency/firmware/devices/dummy/flash.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package devices
+package dummy
```
mhutchinson, line 15: go vet wants `DummyDevice` to be called `Device` to avoid stuttering now its in this package.

AlCutter: Done

PR #148, opened 2020-11-20
Monitor changes

yogeshbdeshpande:


binary_transparency/firmware/cmd/ft_monitor/main.go
```diff
@@ -116,13 +121,27 @@ func main() {
...
 			}
+
 			// Verify Image Hash from log Manifest matches the actual image hash
 			h := sha512.Sum512(image)
 			if !bytes.Equal(h[:], meta.FirmwareImageSHA512) {
 				glog.Warningf("downloaded image does not match SHA512 in metadata (%x != %x)", h[:], meta.FirmwareImageSHA512)
 				continue
 			}
 			glog.V(1).Infof("Image Hash Verified for image at leaf index %d", manifest.LeafIndex)
+
```
AlCutter, line 132: Travis integration test is complaining about formatting:

```
binary_transparency/firmware/cmd/ft_monitor/main.go:133: File is not `goimports`-ed (goimports)
```

binary_transparency/firmware/cmd/ft_monitor/main.go
```diff
@@ -54,6 +57,8 @@ func main() {
 		glog.Exitf("Failed to parse FT log URL: %q", err)
 	}
 
+	ftKeyWords := (*keyWord)
```
AlCutter, line 60: no need to alias for now (we're not modifying/deriving from the value), just use the dereferenced flag later.

binary_transparency/firmware/cmd/ft_monitor/main.go
```diff
@@ -116,13 +121,27 @@ func main() {
...
 			h := sha512.Sum512(image)
 			if !bytes.Equal(h[:], meta.FirmwareImageSHA512) {
 				glog.Warningf("downloaded image does not match SHA512 in metadata (%x != %x)", h[:], meta.FirmwareImageSHA512)
 				continue
 			}
 			glog.V(1).Infof("Image Hash Verified for image at leaf index %d", manifest.LeafIndex)
+
+
+			//Search for specific keywords inside firmware image
+			for _, keyword := range strings.Fields(ftKeyWords) {
```
AlCutter, line 135: just a thought - you could do a bunch of this work just once at line 68:

```go
var ftKeywords []*regex.Regexp
for _, k := range strings.Fields(*keyWord) {
  ftKeywords = append(ftKeywords, regex.MustCompile(k))
}
```

then down here becomes just

```go
for _, m := range ftKeywords {
  if m.Match(image) {
     // ARGH!
  }
}
```

It might also be worth considering explicitly calling out the flag as being a regex, and potentially just letting the user specify things like `\b(malware|virus|malicious|...)\b` in the flag directly - the regex compiler will be able to make the searches sublinear then.

AlCutter (commented):
A couple of small suggestions inline

binary_transparency/firmware/cmd/ft_monitor/main.go
yogeshbdeshpande: Thanks, will check, what is the issue here!

binary_transparency/firmware/cmd/ft_monitor/main.go
yogeshbdeshpande: Agree can do this change!

binary_transparency/firmware/cmd/ft_monitor/main.go
yogeshbdeshpande: Agree on your 1st Comment, will consider better aligning code, the way you indicated. 
However, I request, we just keep the input for user to be simple a string with bunch of keywords, which user wants to search for, **like below.** He should not care to construct something special like regular expression!

"XYZ PQR LMQ"

AlCutter: Sure thing

yogeshbdeshpande: All comments completed and done. Please review and merge!

AlCutter (approved):

PR #151, opened 2020-11-23
Send a few bits of information through LEDs

AlCutter:


binary_transparency/firmware/devices/usb_armory/bootloader/led.go
```diff
@@ -0,0 +1,43 @@
...
+func blinkOnOff(led string, on, off time.Duration) {
+	usbarmory.LED(led, true)
+	<-time.After(on)
+	usbarmory.LED(led, false)
+	<-time.After(off)
+}
+
+// haltAndCatchFire signals for help and never returns.
+func haltAndCatchFire(msg string, code int) {
+	fmt.Println(msg)
```
mhutchinson, line 22: Maybe it doesn't matter, but I think this will go to stdout and I would expect stderr

mhutchinson (approved):

binary_transparency/firmware/devices/usb_armory/bootloader/led.go
AlCutter: I suspect it doesn't really matter here - we're on a tty on the end of a serial console at this point.

PR #154, opened 2020-11-24
Ignore generated bootloader files

AlCutter:


binary_transparency/firmware/devices/usb_armory/bootloader/.gitignore
```diff
@@ -0,0 +1,4 @@
+armory-boot
```
mhutchinson, line 1: What is generating these files? What worries me is that these files will be generated in `pwd` whenever you run `go build` or run the tool, and thus they could appear anywhere in the tree. I much prefer approaches like `**/*.genfilesuffix` or `bin/**` at the root level of the repo. But maybe this is advanced tech that is commonly used and I'm not up to speed?

AlCutter: These are created by the `Makefile` in this dir, so they should always land here.
The Makefile could probably be changed to put them somewhere else, but we still have the issue of the golang bins.

I don't really mind what we do, I was just feeling a bit of burn with the clutter in the `git status`

mhutchinson (approved):
Meh whatever. We can always remove it again.

AlCutter:
> Meh whatever. We can always remove it again.

For sure, but if you feel strongly about it I can also just put `.gitignore` in my own repo, which itself contains a '.gitignore' entry, and that'll fix my issue.

binary_transparency/firmware/devices/usb_armory/bootloader/.gitignore
mhutchinson: If the Makefile will always put the files here with these names then it's reasonable.

PR #155, opened 2020-11-24
Create codeql-analysis.yml

AlCutter:


mhutchinson (approved):
I have no way of testing this with my eyes, but it looks pretty vanilla apart from the `go` selection so LGTM.

PR #156, opened 2020-11-24
Tighten bundle verification a bit

AlCutter:
This makes it harder to not compare image hash/firmware measurement when verifying the bundle.


mhutchinson (approved):
Nice

PR #161, opened 2020-11-26
Support limiting the number of entries to be committed to by the Map

mhutchinson:
This allows maps with the same number of entries to be computed by different parties to compare equality of the root commitment. Also fixed some parameter checking which wasn't working; cleaning the filepath is more error prone than just using the raw input as it swaps empty input for '.'

### DIFF
--- a/experimental/batchmap/sumdb/build/map.go
+++ b/experimental/batchmap/sumdb/build/map.go
@@ -67,11 +67,11 @@ func main() {
 	flag.Parse()
 	beam.Init()
 
-	if *sumDB == "" {
+	if len(*sumDB) == 0 {
 		glog.Exitf("No sum_db provided")
 	}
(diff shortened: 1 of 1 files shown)


experimental/batchmap/sumdb/build/map.go
```diff
@@ -67,28 +67,24 @@ func main() {
 	flag.Parse()
 	beam.Init()
 
-	sumDB := filepath.Clean(*sumDB)
-	if sumDB == "" {
+	if *sumDB == "" {
```
AlCutter, line 70: general nit: use len(string) rather than `... == ""` as it copes with `nil` correctly too.

AlCutter (approved):

AlCutter (approved):

PR #163, opened 2020-11-26
Tick some boxes!

AlCutter:


mhutchinson (approved):
This is the first box ticking exercise I've smiled at in a long time!

PR #164, opened 2020-11-26
Tidy design doc a bit

AlCutter:
- Remove old demo script (we have the new, fuller, one in the FT README)
- Explicitly map the roles/actors to the software from the demo they use

### DIFF
--- a/binary_transparency/firmware/docs/design/README.md
+++ b/binary_transparency/firmware/docs/design/README.md
@@ -138,9 +138,17 @@ here:
 **Firmware vendor**:
 
 Uses the [`publisher`](/binary_transparency/firmware/cmd/publisher) to publish
-firmware metadata & images to the log, and create an "OTA" update bundle.
+firmware metadata & images to the log, and create an "OTA" update bundle:
 
 * Claimant<sup>FIRMWARE</sup>
+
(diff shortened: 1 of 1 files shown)


SimonFrost-Arm (approved):

binary_transparency/firmware/docs/design/README.md
```diff
@@ -131,27 +131,62 @@ following diagram:
...
+For clarity, the mapping of actors to the claimant model roles, along with
+software provided by the demo used to fulfil those roles, are listed explicitly
+here:
 
 **Firmware vendor**:
+
+Uses the [`publisher`](/binary_transparency/firmware/cmd/publisher) to publish
+firmware metadata & images to the log, and create an "OTA" update bundle.
+
 * Claimant<sup>FIRMWARE</sup>
```
mhutchinson, line 143: We've missed `Verifier(FIRMWARE)` for the "I should check my key material is only being applied when I expected" use case

mhutchinson (approved):

binary_transparency/firmware/docs/design/README.md
AlCutter: Good spot, thanks!
PTAL.

PR #165, opened 2020-11-27
Merge upstream bootloader

AlCutter:
- Pulls in upstream updates to allow booting unikernels.
- Backs out the LED signalling changes to keep the diff from upstream as small as possible.


mhutchinson (approved):
LGTM based on trust that this is the same as upstream :-)

PR #166, opened 2020-11-27
Use a database for storing the map tiles instead of raw files

mhutchinson:
This has a number of benefits:
* Avoids the risk of blowing inodes on filesystems
* Heads in the right direction for a real deployent (though a different SQL DB would be used than sqlite for distributed work)
* Allows extra metadata to be stored in the future, e.g. the source info such as SumDB checkpoint, number of entries
* Supports incremental update because the previous tiles can be read and then batchmap.Update used instead of creating a new Map every time


experimental/batchmap/sumdb/build/map.go
```diff
@@ -35,22 +33,27 @@ import (
...
 
+	"github.com/google/trillian-examples/experimental/batchmap/sumdb/mapdb"
+
 	_ "github.com/mattn/go-sqlite3"
 )
 
 const hash = crypto.SHA512_256
 
+// Max number of map rows to write at once. This was plucked from nowhere, change if needed.
+const batchSize = 250
```
AlCutter, line 44: Given the provenance, would it be a good idea to make this a flag?

experimental/batchmap/sumdb/build/map.go
```diff
@@ -108,6 +116,28 @@ func sourceFromFlags(s beam.Scope) beam.PCollection {
 	return databaseio.Query(s, "sqlite3", *sumDB, fmt.Sprintf("SELECT * FROM leafMetadata WHERE id < %d", *count), reflect.TypeOf(Metadata{}))
 }
 
+func sinkFromFlags() (*mapdb.TileDB, int, error) {
+	tiledb, err := mapdb.NewTileDB(*mapDB)
+	if err != nil {
+		return nil, 0, fmt.Errorf("Failed to open map DB at %q: %v", *mapDB, err)
```
AlCutter, line 122: nit: lower case initial letter for errors

experimental/batchmap/sumdb/mapdb/tiledb.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/google/trillian/experimental/batchmap/tilepb"
+)
+
+type NoRevisionsFound = error
```
AlCutter, line 26: nit: Should probably have a docstring on here given it's exported.ni

AlCutter (approved):

PR #172, opened 2020-11-30
Store and check Golden Checkpoints

mhutchinson:
* Checkpoints for successfully validated trees are stored in a new table in the database
* Clone tool now checks that the local tree is consistent with the previous Golden Checkpoint
* Short-circuit the clone tool if the remote SumDB is reporting the same size as the Golden Checkpoint

### DIFF
--- a/sumdbaudit/audit/database.go
+++ b/sumdbaudit/audit/database.go
@@ -19,6 +19,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"time"
 )
 
 // NoDataFound is returned when the DB appears valid but has no data in it.
@@ -54,6 +55,9 @@ func (d *Database) Init() error {
 	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS tiles (height INTEGER, level INTEGER, offset INTEGER, hashes BLOB, PRIMARY KEY (height, level, offset))"); err != nil {
 		return err
 	}
+	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS checkpoints (datetime TIMESTAMP PRIMARY KEY, checkpoint BLOB)"); err != nil {
+		return err
+	}
 	_, err := d.db.Exec("CREATE TABLE IF NOT EXISTS leafMetadata (id INTEGER PRIMARY KEY, module TEXT, version TEXT, repohash TEXT, modhash TEXT)")
 	return err
 }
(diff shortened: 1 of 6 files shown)


sumdbaudit/audit/database.go
```diff
@@ -70,6 +74,29 @@ func (d *Database) Head() (int64, error) {
 	return 0, NoDataFound(errors.New("no data found"))
 }
 
+// GoldenCheckpoint gets the latest checkpoint, using the provided function to parse the note data.
+func (d *Database) GoldenCheckpoint(parse func([]byte) (*Checkpoint, error)) (*Checkpoint, error) {
+	var datetime sql.NullTime
+	var data []byte
+	if err := d.db.QueryRow("SELECT datetime, checkpoint FROM checkpoints ORDER BY datetime DESC LIMIT 1").Scan(&datetime, &data); err != nil {
+		return nil, fmt.Errorf("failed to get max revision: %v", err)
```
AlCutter, line 82: `failed to get latest checkpoint`

AlCutter: (also can use %w)

sumdbaudit/audit/service.go
```diff
@@ -62,10 +62,19 @@ func NewService(localDB *Database, sumDB *SumDBClient, height int) *Service {
 	}
 }
 
+// GoldenCheckpoint gets the previously checked Checkpoint, or nil if not found.
+func (s *Service) GoldenCheckpoint(ctx context.Context) *Checkpoint {
+	golden, err := s.localDB.GoldenCheckpoint(s.sumDB.ParseCheckpointNote)
+	if err != nil {
+		return nil
```
AlCutter, line 69: This seems to silently drop the error, could it return it or at least log it?

sumdbaudit/audit/service.go
```diff
@@ -184,10 +193,107 @@ func (s *Service) HashTiles(ctx context.Context, checkpoint *tlog.Tree) error {
 	return nil
 }
 
+// CheckConsistency checks that the downloaded and hashed tiles are consistent with
+// the previously fetched Golden Checkpoint. Any previously fetched tiles are checked
+// by HashTiles, so this is only really checking that the stragglers that were
+// transiently acquired when the Golden Checkpoint was committed to were actually
+// kept the same and committed to when their tile was completed.
```
AlCutter, line 200: I find it quite hard to parse this sentence, can you refactor a bit?

sumdbaudit/audit/service.go
```diff
@@ -184,10 +193,107 @@ func (s *Service) HashTiles(ctx context.Context, checkpoint *tlog.Tree) error {
...
+
+	root, err := logRange.GetRootHash(nil)
+	if err != nil {
+		return fmt.Errorf("failed to get root hash: %v", err)
+	}
+	var rootHash tlog.Hash
+	copy(rootHash[:], root)
+	if rootHash != checkpoint.Hash {
+		return fmt.Errorf("log root mismatch at tree size %d; calculated %x, SumDB says %x", checkpoint.N, root, checkpoint.Hash[:])
+	}
```
AlCutter, line 289: There's a bit of overlap at the top & tail of this method with the `CheckConsistency` method, is it possible/would it make sense to factor out and reuse?

sumdbaudit/audit/service.go
mhutchinson: Logged it for now. I'm eyeing up a refactor of this whole service's API as its a bit clunky at the moment. Ideally it would have something like a `Sync` or `Mirror` method that gets it up to the latest head.

sumdbaudit/audit/service.go
mhutchinson: Refactored the straggler processing into this method via a function to get them.

AlCutter (approved):
nice!

sumdbaudit/audit/service.go
```diff
@@ -222,39 +295,41 @@ func (s *Service) CheckRootHash(ctx context.Context, checkpoint *tlog.Tree) erro
...
-		sHash := tlog.RecordHash(s)
-		logRange.Append(sHash[:], nil)
+	// Now add all of the stragglers to the logRange.
+	stragglersCount := int(uint64(cp.N) - logRange.End())
+	if stragglersCount > 0 {
+		stragglers, err := getStragglers(stragglersCount)
+		if err != nil {
+			return fmt.Errorf("failed to get stragglers: %w", err)
+		}
+		if err != nil {
```
AlCutter, line 305: Ah, copy pasta error here?

mhutchinson: Something like that :-)

Issue #82, opened 2020-06-18
Provide simple example


lukehinds:
The current examples are really good, but I think it would be useful to have a 'hello world' level example for those new to trillian. 

Such as start a log server, signer, make a simple entry, retrieve the entry.

smeiklej:
i've just merged a new example (#173) that is hopefully doing exactly this - please check it out and let me know what you think!

lukehinds:
thanks @smeiklej , will take a look

mhutchinson:
Luke I'm going to close this. Please reopen this bug (or better yet, raise a new one) with any specific comments on the new example.

PR #173, opened 2020-12-01
hello world example

smeiklej:
A minimal example to demonstrate how to create and use a Trillian personality - addresses issue #82 .


mhutchinson:
You'll need to get this working in Travis. There should be example scripts that bring up/tear down Trillian in Travis in the binary_transparency example.

smeiklej:
@googlebot I signed it!

mhutchinson:
Sarah can you rebase this and we can run the Travis/GCB tests again and try to get this in.

helloworld/README.md
```diff
@@ -0,0 +1,33 @@
+# helloworld
+This is a very basic example of how to use
+[Trillian](http://github.com/google/trillian) to implement the operations
+needed for a verifiable log.  It intentionally avoids
+defining any interface between the client and personality, leaving only the
+gRPC interface between the personality and the (Trillian) log.  In a real
```
mhutchinson, line 6: If this is supposed to be The Simplest Thing, we might want to think about how we introduce client / personality / Trillian. If the terms and relationships are well defined elsewhere we can link to that. Some kind of primer would be handy.

helloworld/README.md
```diff
@@ -0,0 +1,33 @@
...
+deployment it would be necessary to define an interface between the client
+and the personality, and thus implement a "real" client on top of the
```
mhutchinson, line 8: This sentence is a little unclear to me (from "and thus" onwards). I think it might be better to replace this with something like ", for example a RESTful API or gRPC".

helloworld/README.md
```diff
@@ -0,0 +1,33 @@
...
+information).  In one terminal, run the following commands:
+
+```
+export GO111MODULE=on
+go run github.com/google/trillian/cmd/trillian_log_server --rpc_endpoint="localhost:50054" --http_endpoint="localhost:50055" &
+go run github.com/google/trillian/cmd/trillian_log_signer --sequencer_interval="1s" --batch_size=500 --rpc_endpoint="localhost:50056" --http_endpoint="localhost:50057" --num_sequencers=1 --force_master &
+# Trillian services are now running
+
+export TREE_ID=$(go run github.com/google/trillian/cmd/createtree --admin_server=localhost:50054)
+```
```
mhutchinson, line 24: Consider (but you're obliged to pursue) the docker instructions. IMO installing docker is easier than setting up and provisioning a DB in this day and age.

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2020 Google LLC
```
mhutchinson, line 1: nit: it's 2021 (though it wasn't when you started this!)

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
...
+	trillian "github.com/google/trillian"
+	"github.com/google/trillian/merkle/logverifier"
+	"github.com/google/trillian/merkle/rfc6962/hasher"
+
+	p "github.com/google/trillian-examples/helloworld/personality"
+)
+
+// Personality is the interface to the the Trillian log.
+type Personality interface {
+	// Append stores an entry in the log.
```
mhutchinson, line 30: ... and blocks until it is included, returning a Checkpoint which commits to it?

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
...
+	Append(context.Context, []byte) p.Chkpt
+
+	// ProveIncl proves the inclusion of a given entry.
```
mhutchinson, line 33: As this is helloworld, I think it's worth spelling it out here that this returns a proof, and the client should review the proof. A beginner uncertainty is where the proof is created / checked etc.

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
...
+	ProveIncl(context.Context, p.Chkpt, []byte) *trillian.Proof
+
+	// UpdateChkpt provides a new checkpoint and proves it is consistent
+	// with an old one.
```
mhutchinson, line 37: Similar to above, I think spelling out that the proof should be checked.

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
...
+
+// A client is a verifier that maintains a checkpoint as state.
+type Client struct {
+	v      logverifier.LogVerifier
+	chkpt  p.Chkpt
+	person Personality
+}
+
+// NewClient creates a new client with an empty checkpoint and a given
+// personality to talk to.
```
mhutchinson, line 49: I think it's worth noting here that in a real world situation, the client should persist its golden checkpoint between runs. Initializing from the empty checkpoint means it can believe a different history of the log on each run.

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
...
+func NewClient(prsn Personality) Client {
+	c := Client{person: prsn}
+	v := logverifier.New(hasher.DefaultHasher)
+	c.v = v
+	var rootHash []byte
+	chkpt := p.Chkpt{LogSize: 0, RootHash: rootHash}
+	c.chkpt = chkpt
+	return c
```
mhutchinson, line 57:
```suggestion
	v := logverifier.New(hasher.DefaultHasher)
	var rootHash []byte
	chkpt := p.Chkpt{LogSize: 0, RootHash: rootHash}
	return Client{
	      person: prsn,
	      v: v,
	      chkpt: chkpt,
	}
```

It's more standard / safe to set all the fields in one construction rather than building it up.

helloworld/client.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// UpdateChkpt allows a client to update its stored checkpoint.
+func (c Client) UpdateChkpt(chkptNew p.Chkpt, pf *trillian.Proof) bool {
+	// If there is no checkpoint then just use this one no matter what.
+	if c.chkpt.LogSize != 0 {
+		// Else make sure this new checkpoint is consistent with the current one.
+		if pf == nil {
+			return false
+		}
```
mhutchinson, line 80: I'd be tempted to drop this. If a client passes in nil and things go bang, they got what they asked for.

helloworld/helloworld_test.go
```diff
@@ -0,0 +1,146 @@
...
+var (
+	trillianAddr = flag.String("trillian", "localhost:50054", "Host:port of Trillian Log RPC server")
+	treeID       = flag.Int64("tree_id", 0, "Tree ID")
+)
+
+// TestAppend appends a random entry to the log and ensures that the
+// checkpoint updates properly (locally on the personality's side).
+func TestAppend(t *testing.T) {
+	if *treeID == 0 {
+		t.Skip("--treeID flag unset, skipping test")
```
mhutchinson, line 37:
```suggestion
		t.Skip("--tree_id flag unset, skipping test")
```

Same throughout the other tests.

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
+// Copyright 2019 Google LLC
```
mhutchinson, line 1: We're going backwards in time!

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
...
+
+type TrillianP struct {
+	l      trillian.TrillianLogClient
+	treeID int64
+}
+
+// NewPersonality creates a new Trillian personality from the flags.
+func NewPersonality(logAddr string, treeID int64) TrillianP {
+	if treeID <= 0 {
+		log.Fatalf("tree_id must be provided and positive, got %d", treeID)
```
mhutchinson, line 47: Now this is demo code so maybe it's fine, but in general errors should be passed up, and the top level `main` gets to decide to `panic` or `log.Fatal` or `glog.Exit` etc. I would have returned the errors from here, but if you're committed to this stylistic choice to keep error handling out of the limelight then cool.

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
...
+	req := trillian.GetLatestSignedLogRootRequest{LogId: p.treeID}
+	resp, err := p.l.GetLatestSignedLogRoot(ctx, &req)
+	if err != nil {
+		panic(err)
+	}
+	root := resp.GetSignedLogRoot()
+	var logRoot tt.LogRootV1
+	if err := logRoot.UnmarshalBinary(root.LogRoot); err != nil {
+		panic(err)
+	}
```
mhutchinson, line 85:
```suggestion
        // Unpack the response and convert it to the local Checkpoint representation
	root := resp.GetSignedLogRoot()
	var logRoot tt.LogRootV1
	if err := logRoot.UnmarshalBinary(root.LogRoot); err != nil {
		panic(err)
	}
```

Prefer to comment logical blocks like this to signal the intent. This allows the example code to be more easily digested and scanned over.

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
...
+	// Now fetch the new checkpoint, keep going until it's there or until
+	// five seconds have passed. (This is a terrible idea in any real situation.)
+	var chkptNew Chkpt
+	for start := time.Now(); time.Since(start) < 5*time.Second; {
+		chkptNew = p.GetChkpt(ctx)
+		if chkpt.LogSize < chkptNew.LogSize {
+			break
+		}
+	}
+	return chkptNew
```
mhutchinson, line 107: I feel uncomfortable that this method can block for a while, and then return a checkpoint which doesn't include the element because it took too long to integrate. I'd prefer that this bombed out (with a panic, to be consistent with elsewhere) in the event that we reach the timeout without getting a commitment to the new value.

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
...
+	leaf := p.formLeaf(entry)
+	// Form the request according to the Trillian API.
+	req := trillian.GetInclusionProofByHashRequest{
+		LogId:    p.treeID,
+		LeafHash: leaf.MerkleLeafHash,
+		TreeSize: chkpt.LogSize}
+	// Process the response.
+	resp, err := p.l.GetInclusionProofByHash(ctx, &req)
+	if err != nil {
+		return nil
```
mhutchinson, line 122: Interesting; why doesn't this panic? If we're taking that approach, I think it's best to be consistent or explain in a comment why we swallowed the error instead of failing.

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
...
+func (p TrillianP) UpdateChkpt(ctx context.Context, chkpt Chkpt) (Chkpt, *trillian.Proof) {
+	// First get the latest checkpoint.
+	chkptNew := p.GetChkpt(ctx)
+	// Now get a consistency proof if one is needed.
+	var pf *trillian.Proof
+	if chkptNew.LogSize > chkpt.LogSize {
+		req := trillian.GetConsistencyProofRequest{
+			LogId:          p.treeID,
+			FirstTreeSize:  chkpt.LogSize,
+			SecondTreeSize: chkptNew.LogSize}
```
mhutchinson, line 138:
```suggestion
			SecondTreeSize: chkptNew.LogSize,
		}
```

Given everything else Go is opinionated about, I'm surprised this isn't one of them.

helloworld/personality/personality.go
```diff
@@ -0,0 +1,146 @@
...
+		resp, err := p.l.GetConsistencyProof(ctx, &req)
+		if err != nil {
+			return chkptNew, nil
```
mhutchinson, line 141: Again, I'd just blow up here. OR, pass the error out. This middle ground means that the contract of this method is unclear. It can return a newer checkpoint without a proof. A noobie could assume that meant "oh, sometimes a new checkpoint doesn't need a proof because it's just totally trusted".

mhutchinson (commented):
Happy to discuss any of these comments. Much of this is style, but that's important in demo code :-)

helloworld/README.md
smeiklej: okay, i've taken a stab at defining clients and personalities.  let me know what you think?

smeiklej (commented):
addressed these comments inline. big change is in error handling - as suggested, errors are now returned by the personality and client and acted on in the main test.

helloworld/client.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+
+	trillian "github.com/google/trillian"
+	"github.com/google/trillian/merkle/logverifier"
+	"github.com/google/trillian/merkle/rfc6962/hasher"
+
+	p "github.com/google/trillian-examples/helloworld/personality"
+)
+
+// Personality is the interface to the the Trillian log.
```
mhutchinson, line 28: nit: this comment is technically true at some level of abstraction, but given that a goal of this example is to clarify the actors I think it's a bit confusing.

helloworld/client.go
```diff
@@ -0,0 +1,94 @@
...
+	Append(context.Context, []byte) (*p.Chkpt, error)
+
+	// ProveIncl proves the inclusion of a given entry.  It is the job of
+	// the client to verify this inclusion proof in order to convince
+	// itself that the entry really has been included in the log.
+	ProveIncl(context.Context, *p.Chkpt, []byte) (*trillian.Proof, error)
+
+	// UpdateChkpt provides a new checkpoint and proves it is consistent
+	// with an old one.  Again, it is the job of the client to verify this
+	// consistency proof.
```
mhutchinson, line 42: A+ comments - thanks!

helloworld/client.go
```diff
@@ -0,0 +1,94 @@
...
+		v:      v,
+		chkpt:  chkpt,
+	}
+}
+
+// VerIncl allows the client to check inclusion of a given entry.
+func (c Client) VerIncl(entry []byte, pf *trillian.Proof) bool {
+	if pf == nil {
+		return false
+	}
```
mhutchinson, line 71: In Go it's not standard to check for nil input. But you can leave it if it you really like.

helloworld/client.go
```diff
@@ -0,0 +1,94 @@
...
+	if err := c.v.VerifyInclusionProof(pf.LeafIndex, c.chkpt.LogSize,
+		pf.Hashes, c.chkpt.RootHash, leafHash); err != nil {
+		return false
+	}
+	return true
+}
+
+// UpdateChkpt allows a client to update its stored checkpoint.
+func (c Client) UpdateChkpt(chkptNew *p.Chkpt, pf *trillian.Proof) bool {
+	// If there is no checkpoint then just use this one no matter what.
```
mhutchinson, line 82: One thing that might be worth calling out for "in the real world" is that the checkpoint would be signed by the personality, and the client should verify the signature before doing anything else with it.

helloworld/helloworld_test.go
```diff
@@ -0,0 +1,176 @@
...
+	if *treeID == 0 {
+		t.Skip("--tree_id flag unset, skipping test")
+	}
+
+	name := "testAppend"
+	t.Run(name, func(t *testing.T) {
+		ctx := context.Background()
+		personality, err := p.NewPersonality(*trillianAddr, *treeID)
+		if err != nil {
+			panic(err)
```
mhutchinson, line 46: t.Fatalf should be used instead of panic in the test code.

mhutchinson (approved):
LGTM with the changes mentioned (these should be of limited enough scope that another review isn't likely to be needed, but feel free to request a look if you think it needs it)

smeiklej (commented):
addressing the latest comments.

PR #176, opened 2020-12-01
Support incremental update of the map

mhutchinson:
This should be much more efficient for small deltas than rebuilding the whole map each time. This approximately completes the demo of the features and expected usage of the batchmap for the time being.


AlCutter (approved):

mhutchinson:
Ran it locally (single machine):

`Updating revision 3 with range [1966080, 1968896)` - took 156s instead of the 1200s it takes to build the whole log at that size.

PR #180, opened 2020-12-02
Tighten bolts on the armory

AlCutter:
- Adds support for distinct firmware image hash, and firmware measurement hashes
- update docs


binary_transparency/firmware/cmd/publisher/publish.go
```diff
@@ -117,13 +122,28 @@ func main() {
...
 
 func createManifestFromFlags() (api.FirmwareMetadata, []byte, error) {
+	var measure func([]byte) ([]byte, error)
+	switch *deviceID {
+	case "armory":
+		measure = usbarmory.ExpectedMeasurement
+	case "dummy":
+		measure = dummy.ExpectedMeasurement
+	default:
+		glog.Exitf("--device must be one of: 'dummy', 'armory'")
```
mhutchinson, line 132: This should return the error from here and `main` should Exit.

binary_transparency/firmware/devices/dummy/rom/rom.go
```diff
@@ -65,13 +65,20 @@ func ResetFromFlags() (Chain, error) {
...
 
-	// TODO(al): measure firmware and check against manifest expected measurement.
-	// HACK: this should use the firmware measurement field, but it'll do for now.
 	fw, err := ioutil.ReadFile(fwFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read firmware: %w", err)
 	}
-	fwHash := sha512.Sum512(fw)
+
+	// measure firmware
```
mhutchinson, line 73: Can this just call out to `dummy.ExpectedMeasurement(fw)`?

mhutchinson (approved):

binary_transparency/firmware/cmd/publisher/publish.go
AlCutter: done, ta

binary_transparency/firmware/devices/dummy/rom/rom.go
AlCutter: Yeah, I had some half baked idea about keeping the generation and measurement of this separate, but your suggestion results in simpler code so I'll do that for now and break out later if need be.

PR #182, opened 2020-12-02
First iteration of Bundle Fix

yogeshbdeshpande:


binary_transparency/firmware/cmd/flash_tool/flash_tool.go
```diff
@@ -145,7 +145,21 @@ func verifyUpdate(up api.UpdatePackage) error {
 	return nil
 }
 
+//checkConsistency checks that an update package is consistent with existing device package
 func checkConsistency(c *client.ReadonlyClient, up api.UpdatePackage, dev devices.Device) error {
-	// TODO(al): implement this
+	dc, err := dev.DeviceCheckpoint()
+	if err != nil {
+		return fmt.Errorf("failed to fetch the device checkpoint: %q", err)
```
AlCutter, line 152: can you change the `%q` to `%w` to wrap the errors?

binary_transparency/firmware/internal/verify/bundle.go
```diff
@@ -74,3 +80,14 @@ func verifyBundle(b api.ProofBundle) (api.FirmwareMetadata, error) {
 	}
 	return fwMeta, nil
 }
+
+// BundleForConsistency checks for consistency between Bundle checkpoint,
+// and the checkpoint stored on the device
+func BundleForConsistency(bc api.LogCheckpoint, dc api.LogCheckpoint, cp [][]byte) error {
```
AlCutter, line 86: Could you move the functionality here into `BundleForUpdate`?

The exported methods in this package are named for their use case, so `verify.BundleForUpdate()` would check everything is correct before the caller would proceed to install the update, similarly `verify.BundleForBoot()` would perform checks applicable at boot time.

The idea is that having just two entry points in here makes it harder for accidents to happen with each newly supported device being added.

AlCutter (commented):
Nice one!
A couple of suggestions inline, and looks like there's a problem with the tests, but will be great to have this functionality added.

binary_transparency/firmware/internal/verify/bundle.go
yogeshbdeshpande: Lets discuss! I agree with the philosophy behind the concept. However, in terms of practicality, consistency proof, requires a lot more information making the function arguments complex.

binary_transparency/firmware/cmd/flash_tool/flash_tool.go
yogeshbdeshpande: ok, agreed will do!

PR #186, opened 2020-12-04
Add 'bad path' tests to integration test

AlCutter:


binary_transparency/firmware/integration/ft_functions.sh
```diff
@@ -90,5 +91,26 @@ ft_stop_test() {
 }
 
 banner() {
-  echo "---[$1]----------------------------------------------"
+  echo -e "\x1b[1m---[\x1b[7m$1\x1b[m]----------------------------------------------\x1b[0m"
 }
+
+EXPECT_FAIL() {
```
mhutchinson, line 97: Can you add a short docstring for this? It's mostly the fact that it takes args that is surprising to me.

mhutchinson (approved):

binary_transparency/firmware/integration/ft_functions.sh
AlCutter: done (if you know a better/more idiomatic way of doing this I'd be happy to change in a follow up!)

PR #187, opened 2020-12-06
Adding first draft of FT Architecture

yogeshbdeshpande:
Please review architecture diagram pushed in xml and svg format.

These diagrams can be viewed using draw.io tool!


mhutchinson (commented):
Thanks Yogesh. This is an unusual review for me in that there's no specific places to comment on the code so I'll need to put all comments in here.

On the diagram itself:
* I think it goes into too much detail about the internals of Trillian. I think the async signer and the different tables is too much detail for what is needed to understand this. I suggest keeping only `Trillian (Log Mode)` box
* Make clearer that Trillian is committing to the metadata, and the CAS is storing the images (I think the CAS part is clear, but the former isn't)
* Personally I would include a stickman for (or near) the Device and Monitor boxes to clarify the multi-party nature of this ecosystem

On the format itself: I think that having the source in the repo is a massive step forward, so thanks for that. I'm not a giant fan of this particular format though as it seems to be a non-standard format for draw.io which uses XML to wrap a base64 encoded compressed SVG file. This means that any tiny change to the image will mean that the whole file changes. I suspect we can live with this for now and there are better things to do with our time than rewrite this in another format, but I'd like us to see if there's a better open format that's more revision-friendly for other diagrams.

AlCutter:
* +1 on the Trillian internals - it's not important here.
* The `update client` is missing, that's important because there is a line from there to the log personality for consistency proofs
* To better reflect the expected deployment of the architecture, it might be a good idea to double up the `ft_monitor` box - one next to the f/w vendor (representing them monitoring for use of their identity), and one next to another stickman which represents 3rd parties looking for other types of claim violation.

It looks like the `svg` file embeds the drawio `xml`, so we could conceivably rename the svg to `<thing>.drawio.svg` and ditch the `xml` entirely, but longer term I too strongly prefer the generated-from-checked-in-source approach of diagramming if possible.

yogeshbdeshpande (commented):
Hi,

After checking, tidied up the same diagram and also incorporated all the comments!

Request please review and merge!

PR #188, opened 2020-12-07
Updated documentation for SumDB audit/clone tool

mhutchinson:
Also added check for duplicates inline within the tool instead of it being a separate stage that the user needed to perform. Added a force flag to prevent short-circuiting, which is useful when making code changes to the validation.

### DIFF
--- a/sumdbaudit/README.md
+++ b/sumdbaudit/README.md
@@ -4,19 +4,58 @@ The clone tool downloads all entries from the
 [Go SumDB](https://blog.golang.org/module-mirror-launch) into a local SQLite
 database, and verifies that the downloaded data matches the log commitment.
 
-This auditor provides an example for how log data can be verifiably cloned, and
+## Background
+This is a quick summary of https://blog.golang.org/module-mirror-launch but is not
+intended to replace this introduction.
+If you have no context on Go SumDB, read that intro first :-)
+
+Go SumDB is a Verifiable Log based on Trillian, which contains entries of the form:
+```
+github.com/google/trillian v1.3.11 h1:pPzJPkK06mvXId1LHEAJxIegGgHzzp/FUnycPYfoCMI=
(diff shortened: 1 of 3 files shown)


sumdbaudit/README.md
```diff
@@ -4,19 +4,58 @@ The clone tool downloads all entries from the
...
+This auditor provides an example for how Log data can be verifiably cloned, and
 demonstrates how this can be used as a basis to verify its
 [Claims](https://github.com/google/trillian/blob/master/docs/claimantmodel/).
-In the Go SumDB case, the auditor needs to check that no module+version appears
-in the log twice, as this would provide a vector for different clients to be
-presented with different expected checksums. 
+The Claims checked by this clone & audit tool are:
+ * SumDB Checkpoints/STHs properly commit to all of the data in the Log
+ * Committed entries are never modified
+ * The Log is append-only
```
AlCutter, line 35: These two are the same aren't they?
Maybe change to:
```
   * The log is append-only; committed entries are never modified
```
If you're after hammering this point home?

sumdbaudit/cli/clone/clone.go
```diff
@@ -84,7 +87,20 @@ func main() {
...
+
+	dups, err := db.Duplicates()
+	if err != nil {
+		glog.Exitf("Duplicates: %v", err)
+	}
+	if len(dups) > 0 {
+		for _, d := range dups {
+			glog.Errorf("%d duplicates found for %s %s", d.Count, d.Module, d.Version)
+		}
+		glog.Exitf("Duplicate entries is a critical error")
```
AlCutter, line 100: -> `Exit()`

AlCutter (approved):
Couple minor nits inline

sumdbaudit/README.md
mhutchinson: Hah, this is actually how I wrote it first (well, with a slash instead of a semicolon but YOLO)

PR #189, opened 2020-12-08
Replace architecture diagram with PlantUML rendered version

AlCutter:


binary_transparency/firmware/docs/design/README.md
```diff
@@ -129,7 +129,55 @@ The design for the demo consists of a number of different entities which play
 the roles described in the claimant model above, these are shown in the
 following diagram:
 
-![overview diagram](./overview.svg)
+![ov](./ov.svg)
+<div style="display: none">
```
mhutchinson, line 133: It doesn't appear that this is being respected by the diff tool or previews, i.e. the following shows this startuml text in the doc.

https://github.com/google/trillian-examples/blob/4d4f715614e35697de076df8c9b775a5507c3a0e/binary_transparency/firmware/docs/design/README.md

binary_transparency/firmware/docs/design/README.md
```diff
@@ -129,7 +129,55 @@ The design for the demo consists of a number of different entities which play
...
+<!--
+This is an embedded PlantUML config for the overview diagram.
+If you change it, be sure to update the generated diagram by running `plantuml -tsvg  README.md`!
+-->
+@startuml ov
```
mhutchinson, line 138: I think overview is a better name than ov. Can we go back to that?

mhutchinson (commented):
This is neat!

binary_transparency/firmware/docs/design/README.md
AlCutter: done

binary_transparency/firmware/docs/design/README.md
AlCutter: yeah, it works in the preview tool locally but not on GH - I'm going to check it in as is and see what happens on the "real" doc - can double back and sort it out afterwards if it's broken there too.

mhutchinson: https://gist.github.com/noamtamim/f11982b28602bd7e604c233fbe9d910f#gistcomment-3277778

mhutchinson (approved):
From a quick search I'm pretty convinced you're going to see the UML in the rendered page. But if you want to go for it and sort it out, I won't stop you :-)

PR #191, opened 2020-12-08
Hide diagram source in a details tag

AlCutter:
I think this is probably the best I can do without pulling the diagram source into another file, unless you have any other ideas?


mhutchinson (approved):
This is some kind of beautiful hack. "It's not a bug, it's a feature!"

PR #195, opened 2020-12-09
Created a mirror service which continually runs

mhutchinson:
The clone.go tool is intended to be a one-shot tool, and a gentle introduction/demo. This tool is intended to be run as a service which can continually keep a database up to date.

Future work might be to support different databases (e.g. MySQL starts to look reasonable for something like this), or to have some documentation on how to set this up as a Linux service.

### DIFF
--- a/sumdbaudit/cli/mirror/mirror.go
+++ b/sumdbaudit/cli/mirror/mirror.go
@@ -16,12 +16,10 @@ package main
 
 import (
 	"context"
+	"flag"
 	"time"
 
-	"flag"
-
 	"github.com/golang/glog"
-
 	"github.com/google/trillian-examples/sumdbaudit/audit"
 	_ "github.com/mattn/go-sqlite3"
 )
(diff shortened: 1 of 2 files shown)


sumdbaudit/cli/mirror/mirror.go
```diff
@@ -0,0 +1,79 @@
...
+package main
+
+import (
+	"context"
+	"time"
+
+	"flag"
+
+	"github.com/golang/glog"
+
```
AlCutter, line 24: couple of spurious blank lines in here?

AlCutter (approved):
just a

small


nit inline

sumdbaudit/cli/mirror/mirror.go
```diff
@@ -0,0 +1,79 @@
...
+			glog.Exitf("failed to get latest checkpoint: %s", err)
+		}
+		golden := s.GoldenCheckpoint(ctx)
+		if golden != nil && golden.N >= head.N {
+			glog.Infof("nothing to do: latest SumDB size is %d and local size is %d", head.N, golden.N)
+		} else {
+			glog.Infof("syncing to latest SumDB size %d", head.N)
+			s.Sync(ctx, head)
+			if *unpack {
+				glog.V(1).Infof("processing metadata", head.N)
```
AlCutter, line 69: `sumdbaudit/cli/mirror/mirror.go:69:20: Infof call has arguments but no formatting directives`

AlCutter (approved):

PR #197, opened 2020-12-10
Added docs for setting up mirror as a service

mhutchinson:
These instructions were tested on a Raspberry Pi. They should work for any Linux distro that uses systemd.

### DIFF
--- a/sumdbaudit/README.md
+++ b/sumdbaudit/README.md
@@ -114,7 +114,8 @@ At the time of writing this setup uses almost 600MB of storage for the database.
 
 If you want to have the `leafMetadata` table populated then you can add an extra argument
 to the service definition.
-In the `ExecStart` line above, add `-unpack` and then restart the `sumdbmirror` service.
+In the `ExecStart` line above, add `-unpack` and then restart the `sumdbmirror` service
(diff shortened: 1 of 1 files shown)


sumdbaudit/README.md
```diff
@@ -63,6 +65,60 @@ connection, and in around 10 minutes on a Raspberry Pi connected over WiFi.
...
+journalctl -u sumdbmirror -f
+```
+
+When the mirror service is sleeping, you will be able to query the local database at
+`/var/cache/sumdb/mirror.db` using the example queries in the next section.
+At the time of writing this setup uses almost 600MB of storage for the database.
+
+If you want to have the `leafMetadata` table populated then you can add an extra argument
+to the service definition.
+In the `ExecStart` line above, add `-unpack` and then restart the `sumdbmirror` service.
```
AlCutter, line 117: Maybe add the systemctl command to restart the service here too?

AlCutter (approved):

sumdbaudit/README.md
mhutchinson: Yeah, good call.

PR #198, opened 2020-12-10
Added a simple witness to the SumDB examples

mhutchinson:
This allows the mirror to be easily upgraded to provide a service which is easy to integrate with from any other tooling.

### DIFF
--- a/sumdbaudit/README.md
+++ b/sumdbaudit/README.md
@@ -5,6 +5,9 @@ This directory contains tools for verifiably creating a local copy of the
 database.
  * `cli/clone` is a one-shot tool to clone the Log at its current size
  * `cli/mirror` is a service which continually clones the Log
+ * `cli/witness` is an HTTP service that uses a local clone of the Log to provide
+   checkpoint validation for other clients. This is a very lightweight way of
+   providing some Gossip solution to detect split views.
 
 ## Background
 This is a quick summary of https://blog.golang.org/module-mirror-launch but is not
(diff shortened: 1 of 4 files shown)


sumdbaudit/README.md
```diff
@@ -119,6 +122,38 @@ In the `ExecStart` line above, add `-unpack` and then restart the `sumdbmirror`
...
+
+Once running, the server will be available for GET requests at the listen address
+given as a commandline parameter.
+
+Some example requests that can be made:
+```bash
+# Simply get the latest golden checkpoint
+curl -i http://localhost:8080/golden
+
+# Validate that the witness is consistent with the Checkpoint your go build tools are using
```
AlCutter, line 150: It would be better to fetch the latest "consistent" STH from the witness, and have the local tooling verify consistency as the witness could be dishonest about the result...
Perhaps the witness could return its latest golden+proof to the supplied STH and then the client can reverify?

sumdbaudit/audit/service.go
```diff
@@ -136,6 +156,52 @@ func (s *Service) ProcessMetadata(ctx context.Context, checkpoint *Checkpoint) e
...
 
+// CheckConsistency checks that the given checkpoint is consistent with the state
+// downloaded and committed to by the current Golden Checkpoint. If the given
+// checkpoint is too recent then a CheckpointTooFresh error will be returned. If
+// the checkpoint is inconsistent with the golden checkpoint then the error will
+// be of type InconsistentCheckpoints.
+func (s *Service) CheckConsistency(ctx context.Context, checkpoint *Checkpoint) error {
+	golden, err := s.localDB.GoldenCheckpoint(s.ParseCheckpointNote)
+	if err != nil {
+		return fmt.Errorf("failed to query for golden checkpoint: %v", err)
```
AlCutter, line 167: `%w` throughout :)

sumdbaudit/cli/witness/witness.go
```diff
@@ -0,0 +1,118 @@
...
+func main() {
+	flag.Parse()
+
+	if len(*db) == 0 {
+		glog.Exit("db must be provided")
+	}
+
+	db, err := audit.NewDatabase(*db)
+	if err != nil {
+		glog.Exitf("failed to open DB: %v", err)
```
AlCutter, line 96: super nit: capital F

sumdbaudit/cli/witness/witness.go
```diff
@@ -0,0 +1,118 @@
...
+	r.HandleFunc(fmt.Sprintf("/golden"), server.getGolden).Methods("GET")
+	r.HandleFunc(fmt.Sprintf("/checkConsistency/{cp}"), server.checkConsistency).Methods("GET")
+	glog.Fatal(http.ListenAndServe(*listenAddr, r))
+}
+
+func parseBase64Param(r *http.Request, name string) ([]byte, error) {
+	v := mux.Vars(r)
+	b, err := base64.URLEncoding.DecodeString(v[name])
+	if err != nil {
+		return nil, fmt.Errorf("%s should be URL-safe base64 (%q)", name, err)
```
AlCutter, line 115: %w

AlCutter:
Looks like this'll need a rebase to pick up the cloudbuild configs.

sumdbaudit/README.md
mhutchinson: I've added notes in the README and TODOs in the code to add such proofs. The README now makes it clear that this witness is only suitable for use where it is trusted and not a general purpose adversarial situation.

This would be great to add but would be a significant amount of work right now. It needs:
 * The Missing Feature to be added where we download partial tiles (OR, accept that this will make an outbound request to the log for every inbound request)
 * An API defining
 * A verifying client
 * Code to synthesize the consistency proof

This is all doable, but I'm going to heavily lean on pragmatism and say that this code still provides some value that is otherwise impossible to realize in the current ecosystem (there is currently no way for 2 parties to check that they are seeing consistent views of the log unless they get the same checkpoint), and thus let's launch and iterate.

AlCutter: SGTM - my only fear was someone "cut'n'pasting" this without understanding the context, and ending shooting themselves in the foot. If we're calling it out in the README as coming with some caveats then that's cool.

AlCutter (approved):
I wonder whether it might be worth popping the note about the witness security into a

> :warning: Note that the witness is missing features (outlined below) in o...

style box?

PR #202, opened 2020-12-11
Add simple build tests for usbarmory bits

AlCutter:


integration/Dockerfile
```diff
@@ -1,23 +1,30 @@
...
+FROM ubuntu as testbase
 
 WORKDIR /testbase
 
 ARG GOFLAGS=""
 ENV GOFLAGS=$GOFLAGS
 ENV GO111MODULE=on
 
-RUN apt-get update && apt-get -y install curl unzip wget default-mysql-client docker-compose lsof xxd
+RUN apt-get update && apt-get -y install curl default-mysql-client docker-compose lsof unzip wget xxd
```
mhutchinson, line 12: I wonder what we're using the mysql-client for?

integration/Dockerfile
```diff
@@ -1,23 +1,30 @@
 # This Dockerfile builds a base image for the CloudBuild integration testing.
 # It just extends a golang/linux base image with a few extra tools which the
```
mhutchinson, line 2: You can drop the golang part of this line now.

mhutchinson (approved):

integration/Dockerfile
AlCutter: Ah, it used to be for the resetDB stuff, but I think that's gone now.

PR #203, opened 2020-12-11
Tidy up device side

AlCutter:
- Refactor to improve device-side verification reuse
- Remove redundant signature and proof checks
- Flatten ProofBundle inside UpdatePackage
- Don't force device to parse ProofBundle twice

### DIFF
--- a/binary_transparency/firmware/api/update_package.go
+++ b/binary_transparency/firmware/api/update_package.go
@@ -19,6 +19,7 @@ type UpdatePackage struct {
 	// FirmwareImage is the actual firmware image itself.
 	FirmwareImage []byte
 	// ProofBundle holds the various artifacts required to validate the firmware image.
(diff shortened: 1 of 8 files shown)


binary_transparency/firmware/api/update_package.go
```diff
@@ -19,7 +19,7 @@ type UpdatePackage struct {
 	// FirmwareImage is the actual firmware image itself.
 	FirmwareImage []byte
 	// ProofBundle holds the various artifacts required to validate the firmware image.
```
mhutchinson, line 21: This comment should be extended to note the format of this byte slice (JSON).

mhutchinson (approved):

binary_transparency/firmware/api/update_package.go
AlCutter: ta, done

PR #207, opened 2020-12-14
Update flash tool to support SD cards for the armory

AlCutter:


binary_transparency/firmware/cmd/flash_tool/flash_tool.go
```diff
@@ -71,13 +73,20 @@ func main() {
...
 	var dev devices.Device
-	dev, err = dummy.NewFromFlags()
+	switch *deviceID {
+	case "armory":
+		dev, err = armory_flash.NewFromFlags()
+	case "dummy":
+		dev, err = dummy.NewFromFlags()
+	default:
+		glog.Exit("--device must be one of: 'dummy', 'armory'")
+	}
```
mhutchinson, line 83: BTW, https://golang.org/pkg/plugin/ would be a nifty solution to loading in these different devices. The flag business might be tricky actually, but worth considering if we want to truly have a generic flash tool.

mhutchinson (approved):

binary_transparency/firmware/cmd/flash_tool/flash_tool.go
AlCutter: Yeah, good call - I'll merge this as is but go and investigate for a follow up PR.

PR #211, opened 2020-12-14
Add some badges

AlCutter:


mhutchinson (approved):
Everyone loves badgers. And badges.

PR #212, opened 2020-12-15
Split cmds out into implementation and skeleton main files

AlCutter:
- Splits all the `cmd` main files into implementation and bare-minimum `main` files.
- Uses the `cmd` implementations to build a golang integration test.

### DIFF
--- a/binary_transparency/firmware/README.md
+++ b/binary_transparency/firmware/README.md
@@ -198,7 +198,7 @@ in the last step).
    > In both of these cases, you can use the `--force` flag on the `flash_tool`.
 
    ```bash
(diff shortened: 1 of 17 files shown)


mhutchinson (approved):
I'm not massively enamoured with lots of packages called `impl` but we can rename later. The github tooling didn't show the diffs very well so I needed to check this locally with:
`git diff master --name-only | grep "impl/"`
and then for each diff with:
`git diff master:binary_transparency/firmware/cmd/publisher/publish.go head:binary_transparency/firmware/cmd/publisher/impl/publish.go`

(fetching the branch at all was done with `gh pr checkout 212`)

AlCutter:
Yeah, I feel similarly about pkg names, I just wanted to push on to get stuck into the integration test work so I rolled with it - we'll definitely come back and rename.

Sorry the diff'ing was a pain :/

PR #219, opened 2020-12-16
Added dockerfiles for mirror and witness

mhutchinson:
Docker Compose script brings them both up and makes them work together. Witness will be serving at :8080.

### DIFF
--- a/sumdbaudit/docker/docker-compose.yml
+++ b/sumdbaudit/docker/docker-compose.yml
@@ -25,10 +25,11 @@ services:
       - sqlite-data:/var/cache/sumdb
     command: [
       "--alsologtostderr",
-      "--db=/var/cache/sumdb/mirror.db"
+      "--db=/var/cache/sumdb/mirror.db",
+      "--listen=:8080"
     ]
     ports:
-     - "8080:8000"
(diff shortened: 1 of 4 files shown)


sumdbaudit/docker/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+      context: ../..
+      dockerfile: ./sumdbaudit/docker/witness/Dockerfile
+    volumes:
+      - sqlite-data:/var/cache/sumdb
+    command: [
+      "--alsologtostderr",
+      "--db=/var/cache/sumdb/mirror.db"
+    ]
+    ports:
+     - "8080:8000"
```
AlCutter, line 31: consider setting the flag on the job to made it listen on `8000` too?
Could be a little confusing with docker doing this PAT?

AlCutter (approved):

sumdbaudit/docker/docker-compose.yml
mhutchinson: I've explicitly added the port address so that they are both listening on 8080 and the docker server is just opening the port.

PR #224, opened 2020-12-21
Support for building logs of versions for a module

mhutchinson:
If `build_version_list` is provided on the command line then in addition to the 2 hashes for each module+version being committed to in the map, the map will further commit to the list of versions for each module. This is done by building a log for each module, in which the leaves are the versions (in the same order they are found in the input log). The log root hash is committed to in the map under a key which is the hash of the module name.

The list of versions is written to the DB as a JSON serialized list of strings. This is enough information for the log to be reconstructed. This makes all versions for a module discoverable and verifiable.


mhutchinson:
This is quite a large change. I can split the refactoring commits out separately if preferred.

I think it is comprehensible as one big change, but if anything is unclear the sequence of individual commits is available and will provide more context.

pphaneuf (approved):

PR #225, opened 2020-12-22
Tool to list verified versions for a module

mhutchinson:
This builds on #224 by adding a tool to make use of the log of versions. It pulls the list of versions, constructs the log for it, and checks that the map commits to this log. By default it prints only release versions, but the `--all` flag can be used to list ephemeral versions too (e.g. `v0.0.0-20191025170756-f12f9191a1a4`).

This PR looks bigger than it is because the map verification code was extracted/refactored, and github isn't showing the diff.

### DIFF
--- a/experimental/batchmap/sumdb/versions/versions.go
+++ b/experimental/batchmap/sumdb/versions/versions.go
@@ -19,6 +19,7 @@ import (
 	"crypto"
 	"flag"
 	"fmt"
+	"regexp"
 
 	"github.com/golang/glog"
 	"golang.org/x/mod/sumdb/tlog"
@@ -37,6 +38,7 @@ var (
 	mapDB        = flag.String("map_db", "", "sqlite DB containing the map tiles.")
 	treeID       = flag.Int64("tree_id", 12345, "The ID of the tree. Used as a salt in hashing.")
 	prefixStrata = flag.Int("prefix_strata", 2, "The number of strata of 8-bit strata before the final strata.")
+	showAll      = flag.Bool("all", false, "Only release versions are shown by default, but this will also show ephemeral versions.")
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


experimental/batchmap/sumdb/versions/versions.go
```diff
@@ -0,0 +1,97 @@
...
+			copy(lHash[:], left)
+			copy(rHash[:], right)
+			thash := tlog.NodeHash(lHash, rHash)
+			return thash[:]
+		},
+	}
+	logRange := rf.NewEmptyRange(0)
+	for _, v := range versions {
+		h := hash.New()
+		h.Write([]byte(v))
```
AlCutter, line 81: Did you consciously decide not to add a domain separation prefix to the preimage here?
(`tlog.NodeHash` does do this for internal nodes, there's a corresponding `tlog.RecordHash` used for leaves in the sumdb)

experimental/batchmap/sumdb/versions/versions.go
```diff
@@ -37,6 +38,7 @@ var (
 	mapDB        = flag.String("map_db", "", "sqlite DB containing the map tiles.")
 	treeID       = flag.Int64("tree_id", 12345, "The ID of the tree. Used as a salt in hashing.")
 	prefixStrata = flag.Int("prefix_strata", 2, "The number of strata of 8-bit strata before the final strata.")
+	showAll      = flag.Bool("all", false, "Only release versions are shown by default, but this will also show ephemeral versions.")
```
AlCutter, line 41: /but this/but setting this flag/?

experimental/batchmap/sumdb/versions/versions.go
```diff
@@ -89,9 +91,19 @@ func main() {
...
 	for _, v := range versions {
-		versionString = fmt.Sprintf("%s\n * %s", versionString, v)
+		if *showAll || releaseRegex.MatchString(v) {
+			versionString = fmt.Sprintf("%s\n * %s", versionString, v)
+		} else {
+			skipped++
+		}
+	}
+	if skipped > 0 {
+		versionString = fmt.Sprintf("%s\n(%d omitted version)", versionString, skipped)
```
AlCutter, line 106: /omitted version/omitted non-release versions/?

experimental/batchmap/sumdb/versions/versions.go
mhutchinson: I can't determine what past Martin did consciously or otherwise. It's a fair comment though. This code is verifying what the map is already generating in `master`. https://github.com/google/trillian-examples/blob/master/experimental/batchmap/sumdb/build/pipeline/log.go#L79

Options:
* Submit as-is, and then follow up to change both places atomically in another PR
* Change that generation code in this PR

AlCutter: Ah, given this is pre-existing then a follow up seems like the best bet - keep the PRs as small as possible.

AlCutter: Worth adding a `TODO` comment?

AlCutter (approved):

experimental/batchmap/sumdb/versions/versions.go
mhutchinson: Done!

PR #226, opened 2020-12-23
Refactor the core pipeline out of main

mhutchinson:
This makes most of the logic:
 * testable
 * reusable with different IO sources

This adds a basic test for the equivalence of creating a map from scratch, vs creating one and updating it.

Where this is heading is supporting other IO sources. In particular, the current implementation relies on SQLite. That's cute for local testing, but it would be great if this example could be deployed into GCP and run on Dataflow. That will require different IO sources.


experimental/batchmap/sumdb/build/pipeline/pipeline.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	endID, golden, err := b.getLogEnd(size)
+	if err != nil {
+		return tiles, InputLogMetadata{}, err
+	}
+
+	startID := provenance.Entries
+	if startID >= endID {
+		return tiles, InputLogMetadata{}, fmt.Errorf("startID >= endID (%d > %d)", startID, endID)
```
AlCutter, line 110: `>=` for names vs `>` for values in the parenthesis (how about `startID (%d) >= endID (%d)`?)

AlCutter (approved):
just a nit

experimental/batchmap/sumdb/build/pipeline/pipeline.go
mhutchinson: Good catch

AlCutter (approved):

PR #228, opened 2021-01-07
Added support for client/server SQL databases

mhutchinson:
Having the data in a client/server database opens the way for running a mirror in GCP which can be used to build a verifiable map in dataflow. This is infeasible when the database is a local file.

This has been tested with MariaDB 10.5.8.

N.B. This contains a breaking DB schema change. TEXT is not a great choice because at least MariaDB will group by this case insensitively, which surprisingly caused "duplicates" to be identified. Using a binary representation for these fields avoids such false positives.

### DIFF
--- a/sumdbaudit/audit/database.go
+++ b/sumdbaudit/audit/database.go
@@ -18,8 +18,20 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"flag"
 	"fmt"
 	"time"
+
+	"github.com/golang/glog"
+
+	// While flags are defined in this file, the drivers can be imported here.
+	_ "github.com/go-sql-driver/mysql"
+	_ "github.com/mattn/go-sqlite3"
+)
+
+var (
+	sqliteFile = flag.String("sqlite_file", "", "database file location for local SumDB instance")
+	mysqlURI   = flag.String("mysql_uri", "", "URL of a MySQL database for the local SumDB instance")
 )
 
 // NoDataFound is returned when the DB appears valid but has no data in it.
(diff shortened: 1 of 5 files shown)


sumdbaudit/audit/database.go
```diff
@@ -43,16 +55,35 @@ type Database struct {
 	db *sql.DB
 }
 
-// NewDatabase creates a Database using a file at the given location.
-// If the file doesn't exist it will be created.
-func NewDatabase(location string) (*Database, error) {
-	db, err := sql.Open("sqlite3", location)
+// NewDatabaseFromFlags creates a database from the flags defined in this file.
+// TODO(mhutchinson): This feels ugly to define flags not in the main method.
```
AlCutter, line 59: An alternative might be to just have a `db_uri` flag and define it to be `<golang_db_name>:<db_string>` and do something like this in here:
```golang
func NewDBFromURI(flag string) (...) {
   p := strings.SplitN(flag, ":", 2)
   dbConn, err := sql.Open(p[0], p[1])
   ...
```
?

AlCutter (approved):
just a suggestion on the DB flags not in main

sumdbaudit/audit/database.go
mhutchinson: This is a pretty solid option. I'll take the option of doing nothing for now (always preferred) and see how this looks after bedding in. There are a number of intersecting concerns here that make it so there's no slam dunk:
 * We're getting away with vanilla SQL here, but it could be that we want to separate these to take advantage of different DB features (i.e. see the changes that were required to "vanilla" SQL to make it work with MariaDB. This could become infeasible or inefficient if it's made more generic in the future) 
 * Avoiding importing stuff we don't need. In particular, the import for sqlite requires cgo which I think is not needed otherwise (if we can get away without cgo then much smaller docker containers are possible).
 * Consistency of flags vs code duplication

I think this current approach strikes a reasonable balance, though optimizing with a heavy weighting for the last point.

PR #229, opened 2021-01-07
Don't reprocess leaf metadata which is already processed

mhutchinson:
This short circuit saves a lot of time for mirrors which are updating frequently.

### DIFF
--- a/sumdbaudit/audit/database.go
+++ b/sumdbaudit/audit/database.go
@@ -183,6 +183,19 @@ func (d *Database) SetLeafMetadata(ctx context.Context, start int64, metadata []
 	return tx.Commit()
 }
 
+// MaxLeafMetadata gets the ID of the last entry in the leafMetadata table.
+// If there is no data, then NoDataFound error is returned.
+func (d *Database) MaxLeafMetadata(ctx context.Context) (int64, error) {
+	var head sql.NullInt64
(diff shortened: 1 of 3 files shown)


sumdbaudit/audit/database.go
```diff
@@ -152,6 +152,19 @@ func (d *Database) SetLeafMetadata(ctx context.Context, start int64, metadata []
...
+// If there is no data, then NoDataFound error is returned.
+func (d *Database) MaxLeafMetadata(ctx context.Context) (int64, error) {
+	var head sql.NullInt64
+	if err := d.db.QueryRow("SELECT MAX(id) AS head FROM leafMetadata").Scan(&head); err != nil {
+		return 0, fmt.Errorf("failed to get max metadata: %v", err)
+	}
+	if head.Valid {
+		return head.Int64, nil
+	}
+	return 0, NoDataFound(errors.New("no data found"))
```
AlCutter, line 165: Could make `NoDataFound` a var since it's only used here (and callers can check `err == ErrNoDataFound` rather than doing a type assertion - it'll save a line at the call site at least :) )

AlCutter (approved):

sumdbaudit/audit/database.go
mhutchinson: I knew there was something that could be done to make this tidier but I've still got too much Java in my blood at times :-) Ta!

AlCutter (approved):

PR #231, opened 2021-01-08
Update the flags in the docker configs

mhutchinson:
This was missed in #228 when the flag was renamed.

The next step will be to change this docker deployment to use a client/server database, but this is an easy fix that should happen first.


mhutchinson:
This is superseded by #233

PR #233, opened 2021-01-11
Docker images use mariadb instead of sqlite

mhutchinson:
This also fixes some inconsistency around how we signal no data found from the database layer. We now always return ErrNoDataFound instead of sometimes exposing the raw sql error.

This adds a custom check in the QueryRow.Scan err block for no data found. This appears to be another inconsistency between sqlite and mariadb.

A large TODO from this is to to write tests for the different DB integrations.

Also did `go mod tidy` as things were getting messy.


mhutchinson:
This supersedes #231

AlCutter (approved):

PR #235, opened 2021-01-12
Personality can be deployed in a single command

mhutchinson:
The personality is now deployed at the same time as Trillian if the docker command in the README is followed.

This requires an extra table of storage for the tree details so that the container can be restarted. The service will now create the tree if it hasn't done so already. This saves quite a lot of human-driven configuration from the start of the demo, which massively streamlines it (and mitigates the risk of anything going wrong from incorrect configuration). Having fewer terminals also lowers the cognitive cost for the audience.

### DIFF
--- a/binary_transparency/firmware/cmd/ft_personality/internal/trees/trees.go
+++ b/binary_transparency/firmware/cmd/ft_personality/internal/trees/trees.go
@@ -44,6 +44,10 @@ func NewTreeStorage(db *sql.DB) *TreeStorage {
 
 // EnsureTree gets the tree for this personality, creating it if necessary.
 // Takes a connection to Trillian to use for initialization.
+// This is only safe in a single-replica deployment. In a real production setup
+// this provisioning would likely be done by a human ahead of time, or if this
+// style of automatic deployment was required then some kind of locking would
+// be required to ensure that only one log was created and used by all frontends.
 func (s *TreeStorage) EnsureTree(ctx context.Context, conn grpc.ClientConnInterface) (*trillian.Tree, error) {
 	if err := s.init(); err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


binary_transparency/firmware/cmd/ft_personality/internal/trees/trees.go
```diff
@@ -0,0 +1,123 @@
...
+// NewTreeStorage sets up a TreeStorage using the given DB.
+func NewTreeStorage(db *sql.DB) *TreeStorage {
+	return &TreeStorage{
+		db: db,
+	}
+}
+
+// EnsureTree gets the tree for this personality, creating it if necessary.
+// Takes a connection to Trillian to use for initialization.
+func (s *TreeStorage) EnsureTree(ctx context.Context, conn grpc.ClientConnInterface) (*trillian.Tree, error) {
```
AlCutter, line 47: Worth a comment here about the raciness of this?

go.sum
```diff
@@ -501,6 +501,8 @@ github.com/mattn/go-runewidth v0.0.6/go.mod h1:H031xJmbD/WCDINGzjvQ9THkh0rPKHF+m
 github.com/mattn/go-runewidth v0.0.7/go.mod h1:H031xJmbD/WCDINGzjvQ9THkh0rPKHF+m2gUSrubnMI=
 github.com/mattn/go-runewidth v0.0.8 h1:3tS41NlGYSmhhe/8fhGRzc+z3AYCw1Fe1WAyLuujKs0=
 github.com/mattn/go-runewidth v0.0.8/go.mod h1:H031xJmbD/WCDINGzjvQ9THkh0rPKHF+m2gUSrubnMI=
+github.com/mattn/go-sqlite3 v1.14.5 h1:1IdxlwTNazvbKJQSxoJ5/9ECbEeaTTyeU7sEAZ5KKTQ=
```
AlCutter, line 504: Was this intended?

AlCutter (approved):
Cool - couple of minor nits/questions inline.

binary_transparency/firmware/cmd/ft_personality/internal/trees/trees.go
mhutchinson: Done.

PR #238, opened 2021-01-15
Update Trillian

AlCutter:
- Updates Trillian dep to newer version
- go mod tidy
- update verifiers/hashers to use new locations in Trillian


mhutchinson (approved):
It would have been ideal to have seen the update first, and then the change second (to confirm that the forward aliases we left did entail no breaking change). However, this is fine and hopefully we have smaller binaries now.

AlCutter:
> It would have been ideal to have seen the update first, and then the change second (to confirm that the forward aliases we left did entail no breaking change). However, this is fine and hopefully we have smaller binaries now.

That was my original intention, unfortunately the linter did not like that the code here was using deprecated methods and caused the CI to bail.  I guess we could've force pushed it, but seemed easier to appease the machine.

PR #240, opened 2021-01-20
Fix go.sum

mhutchinson:
If we don't do this now it will interfere with all the PRs and cause merging work


AlCutter (approved):

mhutchinson:
Bah, this has passed in Travis but hasn't propagated to github. Time to find the poking stick.

PR #243, opened 2021-01-26
Construct a map of logs from FT log

mhutchinson:
The map is keyed by device ID, and the value is a log which contains the firmware revisions for that device. The log is stored in the database as a flat JSON list of leaf values, and will need to be rehydrated (i.e. the merkle tree recalculated) each time it is used. This puts a limit on how big the sub-logs can be, but that limit is much higher than the cardinalities we'll use for demos.

This requires that the `mysql` docker container has a port exposed outside of the docker network so that the Beam pipeline can access the DB to source the log data.

This pipeline is something of a strawman. By starting with a map of logs, this demonstrates the most complicated map construction and thus all the raw material should be here to allow developers with less experience with Beam to mould this to take whatever shape of Map we decide is required for a convincing demo.

### DIFF
--- a/binary_transparency/firmware/README.md
+++ b/binary_transparency/firmware/README.md
@@ -339,3 +339,29 @@ go run cmd/publisher/publish.go --logtostderr --v=2 --timestamp="2020-10-10T23:0
 >
 > Anybody else running a monitor also knows that malicious firmware has been
 > logged and can raise the alarm.
+
+Generating a Verifiable Map
+---------------------------
+
+There are some pre-requisites to running this part of the demo:
+ 1. Have set up the environment and successfully run the [Trillian batchmap demo](https://github.com/google/trillian/tree/master/experimental/batchmap)
+ 2. Keep the Python Portable Beam Runner listening on port `8099`
+ 3. Have deployed the FT demo above using Docker, and have some firmware in there
+
+The following command will generate a verifiable map:
+
+* `go run ./cmd/ftmap --alsologtostderr --v=2 --runner=universal --endpoint=localhost:8099 --environment_type=LOOPBACK --map_db ~/ftmap.db --trillian_mysql="test:zaphod@tcp(127.0.0.1:3336)/test"`
+
+TODO(mhutchinson): Enhance the demo to show how devices can use this map to be convinced of useful aggregations across the log.
+
+Developer Notes
+---------------
+
+This section is not intended for a general audience.
+The intended audience is developers of FT that have a deployment and need to inspect or modify it.
+
+#### Docker
+
+Connecting to the Trillian database:
+
+`docker run -it --network deployment_default --rm mariadb mysql -hdeployment_mysql_1 -utest -pzaphod test`
--- a/binary_transparency/firmware/README.md
+++ b/binary_transparency/firmware/README.md
@@ -340,20 +340,6 @@ go run cmd/publisher/publish.go --logtostderr --v=2 --timestamp="2020-10-10T23:0
 > Anybody else running a monitor also knows that malicious firmware has been
 > logged and can raise the alarm.
 
-Generating a Verifiable Map
----------------------------
-
-There are some pre-requisites to running this part of the demo:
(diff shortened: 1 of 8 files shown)


binary_transparency/firmware/README.md
```diff
@@ -339,3 +339,29 @@ go run cmd/publisher/publish.go --logtostderr --v=2 --timestamp="2020-10-10T23:0
 >
 > Anybody else running a monitor also knows that malicious firmware has been
 > logged and can raise the alarm.
+
+Generating a Verifiable Map
+---------------------------
+
```
AlCutter, line 345: Maybe we should more strongly flag this section as researchy/WIP for the moment?

binary_transparency/firmware/cmd/ftmap/map.go
```diff
@@ -0,0 +1,226 @@
...
+	beam.RegisterType(reflect.TypeOf((*logToDBRowFn)(nil)).Elem())
+}
+
+func main() {
+	flag.Parse()
+
+	// Connect to where we will read from and write to.
+	sumDB, err := newTrillianDBFromFlags()
+	if err != nil {
+		glog.Exitf("Failed to initialize from local SumDB: %v", err)
```
AlCutter, line 61: SumDB -> ftLogDB(?) (few other sumdb references dotted around)

binary_transparency/firmware/internal/ftmap/log.go
```diff
@@ -0,0 +1,121 @@
...
+type moduleLogHashFn struct {
+	TreeID int64
+
+	rf *compact.RangeFactory
+}
+
+func (fn *moduleLogHashFn) Setup() {
+	fn.rf = &compact.RangeFactory{
+		Hash: func(left, right []byte) []byte {
+			// There is no particular need for using this hash function, but it was convenient.
```
AlCutter, line 67: I wonder if it's worth just having our own defined locally to avoid confusion from the dep on golang sumdb internals?

binary_transparency/firmware/internal/ftmap/log.go
```diff
@@ -0,0 +1,121 @@
...
+	logRoot, err := logRange.GetRootHash(nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create log for %q: %v", log.DeviceID, err)
+	}
+	h := crypto.SHA512_256.New()
+	h.Write([]byte(log.DeviceID))
+	logKey := h.Sum(nil)
+
+	return &batchmap.Entry{
+		HashKey:   h.Sum(nil),
```
AlCutter, line 94: worth having this be `logKey` to be clear that it's identical?

binary_transparency/firmware/internal/ftmap/mapdb.go
```diff
@@ -0,0 +1,122 @@
...
+
+// NoRevisionsFound is returned when the DB appears valid but has no revisions in it.
+type NoRevisionsFound = error
+
+// MapDB provides read/write access to the generated Map tiles.
+type MapDB struct {
+	db *sql.DB
+}
+
+// NewMapDB creates a MapDB using a file at the given location.
```
AlCutter, line 35: /using/in/ ?

AlCutter (approved):
Few nits inline, but this seems like a good base to experiment with.

binary_transparency/firmware/internal/ftmap/log.go
mhutchinson: Yeah, good call.

binary_transparency/firmware/internal/ftmap/log.go
mhutchinson: Yeah not sure how that ended up looking like that.

binary_transparency/firmware/internal/ftmap/mapdb.go
mhutchinson: I'll leave as "using". It needs to work for both writes and reads. I think "using a file" makes sense. This will need to be replaced with a client/server compatible version to really exercise Dataflow down the line anyway.

PR #245, opened 2021-02-03
First revision of interim FT Witness changes

yogeshbdeshpande:
[WIP] FT Witness work in progress.

### DIFF
--- a/binary_transparency/firmware/api/witness.go
+++ b/binary_transparency/firmware/api/witness.go
@@ -16,13 +16,8 @@ package api
 
 const (
 	//WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
-	WitnessGetCheckpoint = "ft/witness/v0/get-w-checkpoint"
+	WitnessGetCheckpoint = "ft/witness/v0/get-checkpoint"
 )
 
-// WCheckpoint commits to the state of the log.
-type WCheckpoint struct {
-	TreeSize uint64
-	RootHash []byte
-	// The number of nanoseconds since the Unix epoch.
-	TimestampNanos uint64
-}
+//Please refer to LogCheckpoint structure in http.go for the CheckPoint returned.
+//Semantics for LogCheckpoint is identical between FT log and Witness
--- a/binary_transparency/firmware/api/witness.go
+++ b/binary_transparency/firmware/api/witness.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC. All Rights Reserved.
+// Copyright 2021 Google LLC. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,9 +15,9 @@
 package api
 
 const (
-	//WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
+	// WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
 	WitnessGetCheckpoint = "ft/witness/v0/get-checkpoint"
 )
 
-//Please refer to LogCheckpoint structure in http.go for the CheckPoint returned.
-//Semantics for LogCheckpoint is identical between FT log and Witness
+// Please refer to LogCheckpoint structure in http.go for the CheckPoint returned.
+// Semantics for LogCheckpoint is identical between FT log and Witness
--- a/binary_transparency/firmware/api/witness.go
+++ b/binary_transparency/firmware/api/witness.go
@@ -1,0 +1,28 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+const (
+	//WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
+	WitnessGetCheckpoint = "ft/witness/v0/get-w-checkpoint"
+)
+
+// WCheckpoint commits to the state of the log.
+type WCheckpoint struct {
+	TreeSize uint64
+	RootHash []byte
+	// The number of nanoseconds since the Unix epoch.
+	TimestampNanos uint64
+}
--- a/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
+++ b/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC. All Rights Reserved.
+// Copyright 2021 Google LLC. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,8 +13,7 @@
 // limitations under the License.
 
 // Package impl is the implementation of the Firmware Transparency witness server.
-// This requires a Trillian instance to be reachable via gRPC and a tree to have
-// been provisioned.
+// This requires a FT Log to be running at a known address.
 package impl
 
 import (
@@ -40,13 +39,13 @@ type WitnessOpts struct {
 	PollInterval time.Duration
 }
 
-//Main kickstarts the witness
+// Main kickstarts the witness
 func Main(ctx context.Context, opts WitnessOpts) error {
 	if len(opts.WSFile) == 0 {
 		return errors.New("Witness Store file is required")
 	}
 
-	ws, err := ws.NewWstorage(opts.WSFile)
+	ws, err := ws.NewStorage(opts.WSFile)
 	if err != nil {
 		return fmt.Errorf("failed to connect witness store: %w", err)
 	}
--- a/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
+++ b/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
@@ -40,7 +40,7 @@ type WitnessOpts struct {
 	PollInterval time.Duration
 }
 
-//Main kick starts the server
+//Main kickstarts the witness
 func Main(ctx context.Context, opts WitnessOpts) error {
 	if len(opts.WSFile) == 0 {
 		return errors.New("Witness Store file is required")
@@ -52,12 +52,12 @@ func Main(ctx context.Context, opts WitnessOpts) error {
 	}
 
 	glog.Infof("Starting FT witness server...")
-	srv := ih.NewServer(ws)
+	witness := ih.NewWitness(ws, opts.FtLogURL, opts.PollInterval)
 	r := mux.NewRouter()
-	srv.RegisterHandlers(r)
+	witness.RegisterHandlers(r)
 
-	// TO DO Check the context parameter in the API correctly.
-	go srv.PollFTLog(ctx, opts.FtLogURL, opts.PollInterval)
+	go witness.Poll(ctx)
+
 	hServer := &http.Server{
 		Addr:    opts.ListenAddr,
 		Handler: r,
--- a/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
+++ b/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
@@ -1,0 +1,74 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package impl is the implementation of the Firmware Transparency witness server.
+// This requires a Trillian instance to be reachable via gRPC and a tree to have
+// been provisioned.
+package impl
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/golang/glog"
+	ih "github.com/google/trillian-examples/binary_transparency/firmware/cmd/ft_witness/internal/http"
+	"github.com/google/trillian-examples/binary_transparency/firmware/cmd/ft_witness/internal/ws"
+	"github.com/gorilla/mux"
+
+	_ "github.com/google/trillian/merkle/rfc6962" // Load hashers
+)
+
+// WitnessOpts encapsulates options for running an FT witness.
+type WitnessOpts struct {
+	ListenAddr   string
+	WSFile       string
+	FtLogURL     string
+	PollInterval time.Duration
+}
+
+//Main kick starts the server
+func Main(ctx context.Context, opts WitnessOpts) error {
+	if len(opts.WSFile) == 0 {
+		return errors.New("Witness Store file is required")
+	}
+
+	ws, err := ws.NewWstorage(opts.WSFile)
+	if err != nil {
+		return fmt.Errorf("failed to connect witness store: %w", err)
+	}
+
+	glog.Infof("Starting FT witness server...")
+	srv := ih.NewServer(ws)
+	r := mux.NewRouter()
+	srv.RegisterHandlers(r)
+
+	// TO DO Check the context parameter in the API correctly.
+	go srv.PollFTLog(ctx, opts.FtLogURL, opts.PollInterval)
+	hServer := &http.Server{
+		Addr:    opts.ListenAddr,
+		Handler: r,
+	}
+	e := make(chan error, 1)
+	go func() {
+		e <- hServer.ListenAndServe()
+		close(e)
+	}()
+	<-ctx.Done()
+	glog.Info("Server shutting down")
+	hServer.Shutdown(ctx)
+	return <-e
+}
--- a/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
+++ b/binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
@@ -51,7 +51,10 @@ func Main(ctx context.Context, opts WitnessOpts) error {
 	}
 
 	glog.Infof("Starting FT witness server...")
-	witness := ih.NewWitness(ws, opts.FtLogURL, opts.PollInterval)
+	witness, err := ih.NewWitness(ws, opts.FtLogURL, opts.PollInterval)
+	if err != nil {
+		return fmt.Errorf("failed to create new witness: %w", err)
+	}
 	r := mux.NewRouter()
 	witness.RegisterHandlers(r)
 
--- a/binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
+++ b/binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
@@ -1,0 +1,143 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package http contains private implementation details for the FirmwareTransparency witness server.
+package http
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/google/trillian-examples/binary_transparency/firmware/api"
+	"github.com/google/trillian-examples/binary_transparency/firmware/internal/client"
+	"github.com/google/trillian-examples/binary_transparency/firmware/internal/verify"
+	"github.com/gorilla/mux"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// WS is the interface to the  Witness Store, for storage of latest checkpoint
+type WS interface {
+	// Store puts the checkpoint into Witness Store
+	StoreCP(wcp api.LogCheckpoint) error
+
+	// Retrieve gets the stored checkpoint.
+	// Must return status code NotFound if no such checkpoint exists.
+	RetrieveCP() (api.LogCheckpoint, error)
+}
+
+// Server is the core state & handler implementation of the FT Witness
+type Server struct {
+	ws WS
+}
+
+// NewServer creates a new server.
+func NewServer(ws WS) *Server {
+	return &Server{
+		ws: ws,
+	}
+}
+
+// getWcheckpoint returns a checkpoint which is registered with witness
+func (s *Server) getWcheckpoint(w http.ResponseWriter, r *http.Request) {
+	checkpoint, err := s.ws.RetrieveCP()
+	if err != nil {
+		http.Error(w, err.Error(), httpStatusForErr(err))
+		return
+	}
+
+	js, err := json.Marshal(checkpoint)
+	if err != nil {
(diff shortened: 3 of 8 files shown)


binary_transparency/firmware/api/witness.go
```diff
@@ -0,0 +1,28 @@
...
+
+package api
+
+const (
+	//WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
+	WitnessGetCheckpoint = "ft/witness/v0/get-w-checkpoint"
+)
+
+// WCheckpoint commits to the state of the log.
+type WCheckpoint struct {
```
mhutchinson, line 23: I'm not sure this is needed, I think a `LogCheckpoint` will be sufficient. We'll want the log signature to be passed around when that is implemented.

binary_transparency/firmware/api/witness.go
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+const (
+	//WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
+	WitnessGetCheckpoint = "ft/witness/v0/get-w-checkpoint"
```
mhutchinson, line 19:
```suggestion
	WitnessGetCheckpoint = "ft/witness/v0/get-checkpoint"
```

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
```diff
@@ -0,0 +1,74 @@
...
+
+// WitnessOpts encapsulates options for running an FT witness.
+type WitnessOpts struct {
+	ListenAddr   string
+	WSFile       string
+	FtLogURL     string
+	PollInterval time.Duration
+}
+
+//Main kick starts the server
```
mhutchinson, line 43:
```suggestion
// Main kickstarts the server
```

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/google/trillian-examples/binary_transparency/firmware/api"
+	"github.com/google/trillian-examples/binary_transparency/firmware/internal/client"
+	"github.com/google/trillian-examples/binary_transparency/firmware/internal/verify"
+	"github.com/gorilla/mux"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// WS is the interface to the  Witness Store, for storage of latest checkpoint
+type WS interface {
```
mhutchinson, line 36: I feel like `WS` is a little opaque. CAS is a known acronym but this isn't. `WitnessStore` isn't too wordy.

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
```diff
@@ -0,0 +1,143 @@
...
+
+// NewServer creates a new server.
+func NewServer(ws WS) *Server {
+	return &Server{
+		ws: ws,
+	}
+}
+
+// getWcheckpoint returns a checkpoint which is registered with witness
+func (s *Server) getWcheckpoint(w http.ResponseWriter, r *http.Request) {
```
mhutchinson, line 58:
```suggestion
func (s *Server) getCheckpoint(w http.ResponseWriter, r *http.Request) {
```

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
```diff
@@ -0,0 +1,143 @@
...
+	// unreachable
+}
+
+// RegisterHandlers registers HTTP handlers for firmware transparency endpoints.
+func (s *Server) RegisterHandlers(r *mux.Router) {
+	r.HandleFunc(fmt.Sprintf("/%s", api.WitnessGetCheckpoint), s.getWcheckpoint).Methods("GET")
+}
+
+// PollFTLog periodically polls the FT log for updating the witness checkpoint.
+func (s *Server) PollFTLog(ctx context.Context, LogURL string, PollInterval time.Duration) error {
```
mhutchinson, line 93: This method stands out as a bit unusual in this context within a class that is nominally a http server. I suggest the following:
 * Rename this from server.go to witness.go and similarly with the `Server` type to `Witness`
 * Add the log URL and interval to the `Witness` struct
 * This method can then be renamed to `StartPolling` or similar

At the moment this method is only really here so that it can borrow the DB connection to write the results to. It's also begging someone to call it multiple times with different log URLs or intervals and bad things will happen.

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	// unreachable
+}
+
+// RegisterHandlers registers HTTP handlers for firmware transparency endpoints.
+func (s *Server) RegisterHandlers(r *mux.Router) {
+	r.HandleFunc(fmt.Sprintf("/%s", api.WitnessGetCheckpoint), s.getWcheckpoint).Methods("GET")
+}
+
+// PollFTLog periodically polls the FT log for updating the witness checkpoint.
```
mhutchinson, line 92: Comment should be clear that this only returns on error (it doesn't start its own polling thread)

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
mhutchinson, line 93:
```suggestion
func (s *Server) PollFTLog(ctx context.Context, logURL string, pollInterval time.Duration) error {
```

mhutchinson (commented):
The shape of this looks broadly as expected. I've put some comments on things that jumped out. Thanks!

binary_transparency/firmware/api/witness.go
yogeshbdeshpande: Agree, I will also add a comment to refer to LogCheckpoint for reference to what is returned and been symentically similar interface

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
yogeshbdeshpande: Done!

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
yogeshbdeshpande: Done!

binary_transparency/firmware/api/witness.go
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+const (
+	//WitnessGetCheckpoint is the path of the URL to get witness checkpoint.
```
mhutchinson, line 18: nit: Comments in Go should have a space between the `//` and the text. This applies throughout the PR but I'll only comment on it this once to save us both a lot of writing and reading!

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
```diff
@@ -0,0 +1,74 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package impl is the implementation of the Firmware Transparency witness server.
+// This requires a Trillian instance to be reachable via gRPC and a tree to have
+// been provisioned.
```
mhutchinson, line 17: This comment about Trillian and trees seems a bit odd. It's technically true, but at this layer in the stack we should just be talking about an FT Log running at a known address.

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
```diff
@@ -0,0 +1,74 @@
...
+	PollInterval time.Duration
+}
+
+//Main kickstarts the witness
+func Main(ctx context.Context, opts WitnessOpts) error {
+	if len(opts.WSFile) == 0 {
+		return errors.New("Witness Store file is required")
+	}
+
+	ws, err := ws.NewWstorage(opts.WSFile)
```
mhutchinson, line 49: This reads oddly to me. I think `NewStorage` is good enough given the package.

binary_transparency/firmware/api/witness.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 Google LLC. All Rights Reserved.
```
mhutchinson, line 1: super nit: the dates in the new files should be 2021.

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return ws, ws.init()
+}
+
+// init creates the file storage
+func (ws *Wstorage) init() error {
+	// Check if the file exists, if not create one
+	f, err := os.OpenFile(ws.fp, os.O_RDWR|os.O_CREATE, os.ModePerm)
+	if err != nil {
+		log.Fatal(err)
```
mhutchinson, line 53: The error handling in this file seems to have been neglected. There are a lot of `log.Fatal` which shouldn't be happening here in the stack.

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if err := f.Close(); err != nil {
+		log.Fatal(err)
+	}
+	return err
+}
+
+// RetrieveCP gets the checkpoint previously stored.
+func (ws *Wstorage) RetrieveCP() (api.LogCheckpoint, error) {
```
mhutchinson, line 89: Remember that this method needs to return `NotFound` as per its API.

Suggest some tests for error conditions would be useful.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// RegisterHandlers registers HTTP handlers for firmware transparency endpoints.
+func (s *Witness) RegisterHandlers(r *mux.Router) {
+	r.HandleFunc(fmt.Sprintf("/%s", api.WitnessGetCheckpoint), s.getCheckpoint).Methods("GET")
+}
+
+// Poll periodically polls the FT log for updating the witness checkpoint.
+func (s *Witness) Poll(ctx context.Context) error {
+
```
mhutchinson, line 98: nit: blank line not needed

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+	ticker := time.NewTicker(s.pollInterval)
+	ftURL, err := url.Parse(s.logURL)
+	if err != nil {
+		return fmt.Errorf("failed to parse FT log URL: %w", err)
+	}
```
mhutchinson, line 103: optional: consider parsing this in `NewWitness`. I prefer to fail fast on obviously bad input.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+	glog.Infof("Polling FT log %q...", ftURL)
+	c := client.ReadonlyClient{LogURL: ftURL}
+	lv := verify.NewLogVerifier()
+	for {
+
```
mhutchinson, line 108: nit: another blank line after an opening brace.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+		wcp, err := s.ws.RetrieveCP()
+		if err != nil {
+			glog.Warningf("Failed to retrieve store logcheckpoint: %q", err)
+			continue
+		}
```
mhutchinson, line 113: On the face of it, this fills me with dread. If there is an error getting the checkpoint, this is going to get stuck in a permaloop spamming the logs. What am I missing?

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+		return fmt.Errorf("failed to parse FT log URL: %w", err)
+	}
+	glog.Infof("Polling FT log %q...", ftURL)
+	c := client.ReadonlyClient{LogURL: ftURL}
+	lv := verify.NewLogVerifier()
+	for {
+
+		wcp, err := s.ws.RetrieveCP()
+		if err != nil {
+			glog.Warningf("Failed to retrieve store logcheckpoint: %q", err)
```
mhutchinson, line 111:
```suggestion
			glog.Warningf("Failed to retrieve stored logcheckpoint: %q", err)
```

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+
+		select {
+		case <-ticker.C:
+			//
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+		cp, err := c.GetCheckpoint()
+		if err != nil {
+			glog.Warningf("Failed to update LogCheckpoint: %q", err)
```
mhutchinson, line 123:
```suggestion
			glog.Warningf("Failed to get LogCheckpoint: %q", err)
```

get or fetch seems more intuitive to me. If I saw "update" in the log I would intuit that it meant writing the update.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+			continue
+		}
+		if cp.TreeSize <= wcp.TreeSize {
+			continue
```
mhutchinson, line 127: I think the case where the new checkpoint is strictly older is worth at least a `glog.V(1).Infof` message. This should never happen in the demo, and if it does the witness will appear to be stuck. It's probably a configuration issue. (obviously this dynamic changes in a deployment where there are multiple log personality frontends running, but while we're not there I'm erring on the side of logging more information rather than less)

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+		glog.V(1).Infof("Got newer checkpoint %s", cp)
+		// Perform consistency check only for non-zero saved witness tree size
+		if wcp.TreeSize != 0 {
+			consistency, err := c.GetConsistencyProof(api.GetConsistencyRequest{From: wcp.TreeSize, To: cp.TreeSize})
+			if err != nil {
+				glog.Warningf("Failed to fetch the Consistency: %q", err)
+				continue
+			}
+			glog.V(1).Infof("Printing the latest Consistency Proof Information")
+			glog.V(1).Infof("Consistency Proof = %x", consistency.Proof)
```
mhutchinson, line 138: nit: I'd downlevel this to `V(2)` - outside of debugging I can't see this proof being displayed in hex making me feel warm and fuzzy inside. The message below about "Consistency proof for tree size verified" meets the need here.

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
yogeshbdeshpande: This does not require addressing due to an earlier comment!!!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,155 @@
...
+			if err := lv.VerifyConsistencyProof(int64(wcp.TreeSize), int64(cp.TreeSize), wcp.RootHash, cp.RootHash, consistency.Proof); err != nil {
+				// Verification of Consistency Proof failed!!
+				glog.Warningf("Failed verification of Consistency proof %q", err)
+				continue
+			}
+			glog.V(1).Infof("Consistency proof for Treesize %d verified", cp.TreeSize)
+		}
+		wcp = *cp
+
+		if s.ws.StoreCP(wcp) != nil {
```
mhutchinson, line 150:
```suggestion
		if s.ws.StoreCP(*cp) != nil {
```

I'm not sure the variable juggling makes this any more readable.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness_test.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/trillian-examples/binary_transparency/firmware/api"
+)
+
+func TestGetWitnessCheckpoint(t *testing.T) {
```
mhutchinson, line 12: Recommend adding a test for some failure cases. The DB not being initialized and failing to read / store would be useful tests.

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/google/trillian-examples/binary_transparency/firmware/api"
+)
+
+const (
+	fileMask = 0755
+)
+
+// Wstorage is a Witness Storage intended for storing witness checkpoints
+// Currently a simple file is used as a storage mechanism
+type Wstorage struct {
```
mhutchinson, line 34:
```suggestion
type Storage struct {
```

throughout

mhutchinson (commented):
Thanks Yogesh. Mostly nits!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: This condition is <= rather than just <. hence it will spit multiple log points every "Fixed Interval" hence I plan to retain the same, as is.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness_test.go
yogeshbdeshpande: Yes, this is WIP..

binary_transparency/firmware/cmd/ft_witness/internal/http/server.go
yogeshbdeshpande: This is complete now

binary_transparency/firmware/api/witness.go
yogeshbdeshpande: Yes, Done everywhere, thanks for the spot!! :)

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
yogeshbdeshpande: Done!

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
yogeshbdeshpande: Agree! Done changes

binary_transparency/firmware/api/witness.go
yogeshbdeshpande: All New files have now 2021 in the HEADER

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
yogeshbdeshpande: log.Fatalf is now replaced with fmt.Errorf and returned from the store!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: DONE!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: Downgraded to V(2) now, but left the comment as is: Nice and helpful in my view!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness_test.go
yogeshbdeshpande: Done, added Failure use case for Witness and Storage Now. Please have a look!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: for now, I prefer to leave it like this!

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
yogeshbdeshpande: Error test done! In case of Empty Store, LogCheckpoint with no values will be returned!

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: Changed from Warning to Fatal so that the loop exits

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse FT log URL: %w", err)
+	}
+	glog.Infof("Polling FT log %q...", ftURL)
+	c := client.ReadonlyClient{LogURL: ftURL}
+	lv := verify.NewLogVerifier()
+	for {
+		wcp, err := s.ws.RetrieveCP()
+		if err != nil {
+			glog.Fatal("Failed to retrieve stored logcheckpoint: %w", err)
```
mhutchinson, line 109: This should return the error and the caller should terminate.

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
```diff
@@ -0,0 +1,106 @@
...
+		return fmt.Errorf("failed to write data to witness db file: %w", err)
+	}
+	if err := f.Close(); err != nil {
+		return fmt.Errorf("failed to close file: %w", err)
+	}
+	return nil
+}
+
+// RetrieveCP gets the checkpoint previously stored.
+func (ws *Storage) RetrieveCP() (api.LogCheckpoint, error) {
```
mhutchinson, line 90: This method concerns me; if this is called before `StoreCP` then it will fail, and not as the contract specifies "Must return status code NotFound if no such checkpoint exists.". Please make sure that a new witness which cannot read the log does not explode if get-checkpoint endpoint is called.

yogeshbdeshpande: If there is no Checkpoint in the FT Log at the moment does it return Error ? I suppose it would return Empty Checkpoint right? Please correct me if I am wrong??

I am not sure what you mean by your last statement ?? Can you please clarify ??

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: Can change this...

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
mhutchinson: The behaviour of the log is not relevant to the point I'm making about this `Storage` object. This class as implemented will not live up to its contract if `RetrieveCP` is called before `StoreCP`. The contract says it must return `NotFound` but there is no code here to handle that case.

```
s := NewStorage(createRandomFilePath())
cp, err := s.RetrieveCP()
```

That's an empty/non-existent file path. I would expect cp to be nil, and err to contain `NotFound` given the contract on the interface you defined. As implemented it will be a generic `fmt.Errorf`.

yogeshbdeshpande: Hi Martin,

Let's discuss this on call tomorrow please. Perhaps we should clarify the "Interface Contract" that if No Checkpoint exists in the store, it will return an Empty CheckPoint with no Error.

 Otherwise, the first time the Poll starts it will always return an Empty store and should continue, rather than bailing out!

As mentioned above happy to discuss on the call.

Thanks,
Yogesh

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
yogeshbdeshpande: Not needed in the new scheme of things!

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
yogeshbdeshpande: As agreed, golden checkpoint retrieved once in the memory at Witness server creation time, Added Mutex for access where two threads can access the gcp and witness store simultaneously!

mhutchinson (approved):

PR #246, opened 2021-02-11
Added a LogFollower class to encapsulate common logic

mhutchinson:
This will need to be written again for the annotator client, so factoring this out now to avoid yet more duplication of critical code.

This implementation uses go channels to improve separation of concerns.

### DIFF
--- a/binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
+++ b/binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
@@ -126,5 +126,4 @@ func Main(ctx context.Context, opts MonitorOpts) error {
 		}
 
 	}
-	return nil
 }
--- a/binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
+++ b/binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
@@ -35,8 +35,6 @@ import (
 	"github.com/golang/glog"
 	"github.com/google/trillian-examples/binary_transparency/firmware/api"
 	"github.com/google/trillian-examples/binary_transparency/firmware/internal/client"
-	"github.com/google/trillian-examples/binary_transparency/firmware/internal/crypto"
-	"github.com/google/trillian-examples/binary_transparency/firmware/internal/verify"
 )
 
 // MatchFunc is the signature of a function which can be called by the monitor
@@ -52,7 +50,6 @@ type MonitorOpts struct {
 }
 
 func Main(ctx context.Context, opts MonitorOpts) error {
-
 	if len(opts.LogURL) == 0 {
 		return errors.New("log URL is required")
 	}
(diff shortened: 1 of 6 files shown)


binary_transparency/firmware/internal/client/follower.go
```diff
@@ -0,0 +1,185 @@
...
+				glog.V(1).Infof("Consistency proof for Treesize %d verified", cp.TreeSize)
+			}
+			f.golden = *cp
+			outc <- *cp
+		}
+	}()
+	return outc, errc
+}
+
+func (f *LogFollower) Entries(ctx context.Context, cpc <-chan api.LogCheckpoint, head uint64) (<-chan LogEntry, <-chan error) {
```
yogeshbdeshpande, line 136: Comment on Entries, similar to Checkpoints will be helpful!

mhutchinson: Yup I was just working on that actually :-) Now done.

binary_transparency/firmware/internal/client/follower.go
```diff
@@ -0,0 +1,185 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
```
yogeshbdeshpande, line 2: should the name be more appropriate as logfollower or something similar?

binary_transparency/firmware/internal/client/follower.go
```diff
@@ -0,0 +1,185 @@
...
+func (f *LogFollower) Checkpoints(ctx context.Context) (<-chan api.LogCheckpoint, <-chan error) {
+	ticker := time.NewTicker(f.pollInterval)
+
+	outc := make(chan api.LogCheckpoint, 1)
+	errc := make(chan error, 1)
+
+	// Push the starting checkpoint into the channel here to kick off the pipeline
+	outc <- f.golden
+
+	go func() {
```
yogeshbdeshpande, line 87: Just a question, mostly for my benefit.
If @line 85, if this is run as a separate thread, via go func, how the calling code inside Monitor or Witness wait or block for a poll ??

mhutchinson: `outc` and `errc` are buffered with a size of 1, which will prevent immediate blocking if only writing a single element. If we didn't buffer `outc` then this method would never get past L85 as currently written.

Calling code blocks on the channel returned by this method until elements are available. While the channel is empty, the code will block and wait for it. This is effectively the same pattern as before, except then channel was `ticker.C` instead of a custom channel with Checkpoints on it.

binary_transparency/firmware/internal/client/follower.go
mhutchinson: I'm happy with the current name as its a) short, and b) I'd be quite happy to stick and other code for followers (e.g. MapFollower) in this file if needed. We can always rename it later if we want. Go generally prefers shorter naming.

binary_transparency/firmware/internal/client/follower.go
yogeshbdeshpande: Sure thanks, just wondering as this breaks the logic into two for loops (one in monitor and one in follower), can there be a race condition where the monitor thread is not complete and a new checkpoint is available on the channel, will the monitor context of checkpoint over-written in this use case?

mhutchinson: Maybe take a look at https://blog.golang.org/pipelines if you'd like to see more about Go concurrency, especially this "pipeline" style of processing.

yogeshbdeshpande: Sure will do!

yogeshbdeshpande (approved):

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
```diff
@@ -62,132 +59,71 @@ func Main(ctx context.Context, opts MonitorOpts) error {
...
-		case <-ticker.C:
-			//
+		case err = <-cperrc:
+			return err
+		case err = <-eerrc:
+			return err
 		case <-ctx.Done():
 			return ctx.Err()
+		default:
 		}
```
AlCutter, line 89: I'm a bit squinty about this - I think the logic of channels/errs and select works, but it's potentially a bit brittle e.g. if someone later decides that sending an informative error down `cperrc` without bailing out of the goroutine and closing the `cpc` channel is ok.

Wondering if it might be better/more readable to do something like this instead? :
```golang

cpc, cperrc := follow.Checkpoints(ctx, opts.PollInterval, latestCP)
ec, eerrc := follow.Entries(ctx, cpc, head)

for {
  select {
     // handle errs
     case entry, ok := <- ec:
       if !ok {
          break
   }
}
```

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
```diff
@@ -83,57 +82,33 @@ func (s *Witness) RegisterHandlers(r *mux.Router) {
...
+	follow := client.NewLogFollower(c)
+
+	glog.Infof("Polling FT log %q...", ftURL)
+	cpc, cperrc := follow.Checkpoints(ctx, s.pollInterval, s.gcp)
 
+	for cp := range cpc {
 		select {
-		case <-ticker.C:
-			//
+		case err = <-cperrc:
```
AlCutter, line 97: same here?

binary_transparency/firmware/internal/crypto/signature.go
```diff
@@ -178,3 +178,14 @@ func (c *Claimant) VerifySignature(stype api.StatementType, stmt []byte, signatu
...
 }
+
+func ClaimantForType(t api.StatementType) (*Claimant, error) {
+	switch t {
+	case api.FirmwareMetadataType:
+		return &Publisher, nil
+	case api.MalwareStatementType:
+		return &AnnotatorMalware, nil
+	default:
+		return nil, fmt.Errorf("Unknown type %v", t)
```
AlCutter, line 189: /Unknown type/unknown claimant type/?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: I don't have the `ok` return, but I think the new version is more consistent and clearer. PTAL.

binary_transparency/firmware/internal/crypto/signature.go
mhutchinson: Done.

binary_transparency/firmware/cmd/ft_witness/internal/http/witness.go
mhutchinson: Ditto above.

AlCutter (approved):

PR #254, opened 2021-02-26
Refactored the FT Map to lay the foundations for processing annotations

mhutchinson:
The key changes here are:
 * Parse the statements, and then partition them into different collections for each logical type
 * Return a Result object instead of a bunch of collections (to which we will be adding more collections to)

Because the collections are pre-filtered in the partition stage, some of the DoFns have been
simplified to return the parsed object directly instead of using the emitter (which was needed
to support returning nothing, if the statement was the wrong type).

### DIFF
--- a/binary_transparency/firmware/internal/ftmap/pipeline.go
+++ b/binary_transparency/firmware/internal/ftmap/pipeline.go
@@ -29,9 +29,13 @@ import (
 )
 
 func init() {
-	beam.RegisterFunction(parseLogEntryFn)
+	beam.RegisterFunction(parseStatementFn)
+	beam.RegisterFunction(parseFirmwareFn)
+	beam.RegisterFunction(partitionFn)
 	beam.RegisterType(reflect.TypeOf((*InputLogMetadata)(nil)).Elem())
 	beam.RegisterType(reflect.TypeOf((*InputLogLeaf)(nil)).Elem())
+	beam.RegisterType(reflect.TypeOf((*PipelineResult)(nil)).Elem())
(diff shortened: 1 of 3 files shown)


yogeshbdeshpande (approved):

binary_transparency/firmware/internal/ftmap/pipeline.go
```diff
@@ -115,29 +148,50 @@ func (b *MapBuilder) getLogEnd(requiredEntries int64) (int64, []byte, error) {
 	return requiredEntries, golden, nil
 }
 
-// firmwareLogEntry is a parsed version of InputLogLeaf.
-type firmwareLogEntry struct {
-	Index    int64
-	Firmware api.FirmwareMetadata
+// partitionFn partitions the statements by type.
+// If you change the implementation here, then the cardinality needs to be
+// changed where this function is applied.
```
AlCutter, line 153: Just a thought, you could have an explicit set of consts for the partitions like below and avoid/technically enforce the "if you ___, then don't forget to ___" :

```golang
const (
 FirmwareMetaPartition = iota
 MalwareStatementPartition
 // Add new partitions here
 ...
 MaxPartitions
)
```

and use `MaxPartitions` in the `beam.Partition` call directly.

AlCutter (approved):
One mini suggestion inline.

binary_transparency/firmware/internal/ftmap/pipeline.go
mhutchinson: This is super neat, and I forgot Go could do enums for some reason. Done!

PR #257, opened 2021-03-04
Updated Trillian and made required changes

mhutchinson:
This is a big one as the interactive/DB-backed Map has been removed from Trillian, which meant that the `registers` and `etherslurp` examples were left without the library they were demonstrating. They could be fixed up to use the new experimental batchmap if this is a direction we want to go in, but I suggest we drop them and re-add them, or we risk falling further behind Trillian while this work is done.

`HStar2` was also removed, so some verification code was updated to use `HStar3` instead.


AlCutter (approved):
Is it worth popping an issue in as a TODO to consider re-adding these at a future date?

mhutchinson:
I've created issue #266 for the demos. I've had fun rebasing and merging these go.{mod,sum} files, but this should be good now.

PR #276, opened 2021-03-29
[serverless] Flesh out sequencing support

AlCutter:


serverless/internal/storage/fs/fs.go
```diff
@@ -102,21 +116,123 @@ func (fs *Storage) LogState() api.LogState {
...
+		}
+
+		// All done!
+		break
+	}
+
+	return nil
+}
+
+// createExcl, creates the named file before writing the data in d to it.
```
mhutchinson, line 196: `createExcl` took me a few turns to parse. I expanded its name to mean "this methods creates an excl.*" and started expanding this to "exclusion" etc first.

serverless/internal/storage/fs/fs.go
```diff
@@ -102,21 +116,123 @@ func (fs *Storage) LogState() api.LogState {
...
+// This method will attempt to silently squash duplicate leaves, but it cannot
+// be guaranteed that no duplicate entries will exist.
 func (fs *Storage) Sequence(leafhash []byte, leaf []byte) error {
-	return errors.New("unimplemented")
+	// 1. Check for dupe leafhash
+	// 2. Write temp file
+	// 3. Hard link temp -> seq file
+	// 4. Optimistically symlink leafhash -> seq file
+
+	// Ensure the leafhash diretory structure is pressent
```
mhutchinson, line 141:
```suggestion
	// Ensure the leafhash directory structure is present
```

serverless/internal/storage/fs/fs.go
```diff
@@ -23,29 +23,35 @@ import (
...
 //  <rootDir>/leaves/aa/bb/cc/ddeeff...
+//  <rootDir>/leaves/pending/aabbccddeeff...
 //  <rootDir>/seq/aa/bb/cc/ddeeff...
 //  <rootDir>/tile/<level>/aa/bb/ccddee...
 //  <rootDir>/state
 //
 // The functions on this struct are not thread-safe.
 type Storage struct {
 	rootDir string
+	nextSeq uint64
```
mhutchinson, line 48: It would help to comment this with the invariant that it will never point ahead of the next sequence. It may however be pointing at an old entry and this won't cause it to stomp on the already sequenced data. In effect, this is a performance improvement rather than anything functional (if this was always set to 0 the sequencing would work, but it would be slow).

serverless/internal/storage/fs/fs.go
```diff
@@ -102,21 +116,123 @@ func (fs *Storage) LogState() api.LogState {
...
+	if err := tmpFile.Close(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ScanSequenced calls the provided function once for each contiguous entry
+// in storage starting at begin.
+// The scan will abort if the function returns an error.
+func (fs *Storage) ScanSequenced(begin uint64, f func(seq uint64, entry []byte) error) (uint64, error) {
```
mhutchinson, line 220: Document the return value. It could be the last index or the number of sequenced entries. (it's the number of sequenced entries, but I only know from reading the impl)

mhutchinson (approved):

serverless/internal/storage/fs/fs.go
AlCutter: -> `createExclusive`

serverless/internal/storage/fs/fs.go
AlCutter: Man, some high quality typing right there, ta!

serverless/internal/storage/fs/fs.go
AlCutter: done

serverless/internal/storage/fs/fs.go
AlCutter: done

PR #277, opened 2021-03-29
[serverless] Flesh out integration implementation

AlCutter:


AlCutter:
rebased

serverless/internal/log/integrate.go
```diff
@@ -17,25 +17,150 @@
...
+	l, err := strconv.ParseUint(p[0], 16, 64)
+	if err != nil {
+		panic(err)
+	}
+	i, err := strconv.ParseUint(p[1], 16, 64)
+	if err != nil {
+		panic(err)
+	}
+	return l, i
 }
```
mhutchinson, line 166: Does it have to go via a string? You can just make a struct to encode the level & index, e.g. https://play.golang.org/p/E0Ef80nkSsY

serverless/internal/log/integrate.go
```diff
@@ -17,25 +17,150 @@
...
+
+	r, err := baseRange.GetRootHash(nil)
+	if err != nil {
+		return fmt.Errorf("invalid log state, unable to recalculate root: %w", err)
+	}
+	glog.Infof("Loaded state with roothash %x", r)
+
+	tiles := make(map[string]*api.Tile)
+
+	visitor := func(id compact.NodeID, hash []byte) {
```
mhutchinson, line 73: Some documentation on this would be useful. There's a lot going on here as this is binding a) the visitor pattern from compact range, b) the in-memory `tiles` map, c) the actual storage.

The upshot of what it's trying to do from what I can see is: populate `tiles` with all the tiles that need to be updated. This creates new tiles if needed, and updates any existing tiles already in the storage. The tiles will not be written back to storage, but will be stored in `tiles` afterwards.

serverless/internal/log/integrate.go
```diff
@@ -17,25 +17,150 @@
...
+			// Update range and set internal nodes
+			newRange.Append(lh, visitor)
+			return nil
+		})
+	if err != nil {
+		return fmt.Errorf("error while integrating: %w", err)
+	}
+	if n == 0 {
+		glog.Infof("Nothing to do.")
+		// Nothing to do, nothing done.
```
mhutchinson, line 116: nit: this comment is a little superfluous!

serverless/internal/log/integrate.go
```diff
@@ -17,25 +17,150 @@
...
+	if err := baseRange.AppendRange(newRange, visitor); err != nil {
+		return fmt.Errorf("failed to merge new range onto existing log: %w", err)
+	}
+
+	newRoot, err := baseRange.GetRootHash(nil)
+	if err != nil {
+		return fmt.Errorf("failed to calculate new root hash: %w", err)
+	}
+	glog.Infof("New log state: size 0x%x hash: %x", baseRange.End(), newRoot)
+
```
mhutchinson, line 129: A comment here marking the transition from computation to storage would be useful. Above this point nothing is written, but there's a lot going on and that's not easy to see at a glance.

serverless/internal/log/integrate.go
```diff
@@ -17,25 +17,150 @@
...
+			return fmt.Errorf("failed to store tile at level %d index %d: %w", l, i, err)
+		}
+	}
+
+	newState := api.LogState{
+		RootHash: newRoot,
+		Size:     baseRange.End(),
+		Hashes:   baseRange.Hashes(),
+	}
+	if err := st.UpdateState(newState); err != nil {
```
mhutchinson, line 142: Somewhere around here I think it would be useful to briefly discuss what happens in the event of a failure during this writing phase. I think it all works out to be OK because a) nobody will be reading newer entries until the `State` is updated (which happens at the end), b) the sequence that was consumed this time can't be changed, and thus the next Integrate write will be idempotent and thus complete the work that was happening here.

mhutchinson (approved):

serverless/internal/log/integrate.go
```diff
@@ -17,25 +17,150 @@
...
 )
 
 // Storage represents the set of functions needed by the log tooling.
 type Storage interface {
+	// GetTile returns the tile at the given level & index.
+	GetTile(level, index, logSize uint64) (*api.Tile, error)
+
+	// StoreTile stores the tile at the given level & index.
+	StoreTile(level, index uint64, tile *api.Tile) error
+
```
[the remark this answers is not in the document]
AlCutter: Yeah, the README has a checkbox for more detailed docs on the design and structure of the storage layout etc.

serverless/internal/log/integrate.go
AlCutter: Lovely suggestion, ta.

serverless/internal/log/integrate.go
AlCutter: I've factored out the cache/visitor and added a bunch of comments throughout.

serverless/internal/log/integrate.go
AlCutter: done

PR #279, opened 2021-03-30
[serverless] Add tile support

AlCutter:


serverless/internal/storage/fs/fs.go
```diff
@@ -246,12 +247,76 @@ func (fs *Storage) ScanSequenced(begin uint64, f func(seq uint64, entry []byte)
...
-	return nil, errors.New("unimplemented")
+	tileSize := storage.PartialTileSize(level, index, logSize)
+	p := filepath.Join(layout.TilePath(fs.rootDir, level, index, tileSize))
+	t, err := ioutil.ReadFile(p)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, fmt.Errorf("failed to read tile at %q: %w", p, err)
+	}
+	if err != nil {
+		return nil, err
+	}
```
mhutchinson, line 258:
```suggestion
	if err != nil {
		if !errors.Is(err, os.ErrNotExist) {
			return nil, fmt.Errorf("failed to read tile at %q: %w", p, err)
	  	}
	 	return nil, err
	}
```

nit suggestion: It reads a little clearer to me if both errors are handled in the same block.

serverless/internal/storage/fs/fs.go
```diff
@@ -246,12 +247,76 @@ func (fs *Storage) ScanSequenced(begin uint64, f func(seq uint64, entry []byte)
...
+	glog.V(2).Infof("StoreTile: level %d index %x ts: %x", level, index, tileSize)
+	if tileSize == 0 || tileSize > 256 {
+		return fmt.Errorf("tileSize %d must be > 0 and <= 256", tileSize)
+	}
+	t, err := json.Marshal(tile)
+	if err != nil {
+		return fmt.Errorf("failed to marshal tile: %w", err)
+	}
+
+	tDir, tFile := layout.TilePath(fs.rootDir, level, index, uint64(tileSize%256))
```
mhutchinson, line 282: Why is the mod 256 needed here? We just asserted it was in range [0, 256).

serverless/internal/storage/fs/fs.go
```diff
@@ -246,12 +247,76 @@ func (fs *Storage) ScanSequenced(begin uint64, f func(seq uint64, entry []byte)
...
+	if err := os.Rename(temp, tPath); err != nil {
+		return fmt.Errorf("failed to rename temporary tile file: %w", err)
+	}
+
+	if tileSize == 256 {
+		partials, err := filepath.Glob(fmt.Sprintf("%s.*", tPath))
+		if err != nil {
+			return fmt.Errorf("failed to list partial tiles for clean up; %w", err)
+		}
+		// Clean up old partial tiles by symlinking them to the new full tile.
```
mhutchinson, line 303: Any reason to symlink instead of hardlink?

mhutchinson (approved):

serverless/internal/storage/fs/fs.go
AlCutter: done

serverless/internal/storage/fs/fs.go
AlCutter: Well, we asserted that it's [1,256] - i.e. not-empty, and not bigger than 2^8.
The logic in `TilePath` is that full tiles are named `ID` and partial tiles named `ID.xx`, so it takes a param with the number of leaves expected for partial tiles and uses that for .xx iff it's > 0.

I've updated the `TilePath` param name and func comment to help make that clearer.

serverless/internal/storage/fs/fs.go
AlCutter: Just thinking ahead a bit - hard links will be interpreted as multiple separate files in some places (e.g. git)

PR #282, opened 2021-03-31
[serverless] Client proof builder & integration test

AlCutter:
- Adds client-side proof building code
- First cut of integration test

### DIFF
--- a/serverless/integration/integration_test.go
+++ b/serverless/integration/integration_test.go
@@ -1,0 +1,99 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


serverless/integration/integration_test.go
```diff
@@ -0,0 +1,99 @@
...
+		// Sequence some leaves:
+		sequenceNLeaves(t, s, lh, leavesPerLoop)
+
+		// Integrate those leaves
+		if err := log.Integrate(s, lh); err != nil {
+			t.Fatalf("Integrate = %v", err)
+		}
+
+		newState := s.LogState()
+		if got, want := newState.Size-state.Size, uint64(leavesPerLoop); got != want {
```
mhutchinson, line 55: I'd expect deduplication to be kicking in and making there be 0 leaves added after the first loop. What am I missing?

mhutchinson (approved):
Worth updating the PR description. As notable as the integration test is the inclusion of the client proof validation.

serverless/integration/integration_test.go
AlCutter: I suspect the duplicate squashing isn't working - good job we have a test now :)
Fix coming in a follow up.

PR #285, opened 2021-04-01
[serverless] Expose sequence numbers to sequencer

AlCutter:


serverless/internal/storage/fs/layout/layout.go
```diff
@@ -44,6 +47,24 @@ func SeqPath(root string, seq uint64) (string, string) {
...
+		return 0, fmt.Errorf("invalid seqPath %q, doesn't start with prefix %q", seqPath, seqDir)
+	}
+	s := strings.TrimPrefix(seqPath, seqDir)
+	if s[0] != os.PathSeparator || s[3] != os.PathSeparator || s[6] != os.PathSeparator || s[9] != os.PathSeparator || s[12] != os.PathSeparator {
+		return 0, fmt.Errorf("seqPath format invalid %q", s)
+	}
+	var b strings.Builder
+	for _, s := range []string{s[1:3], s[4:6], s[7:9], s[10:12], s[13:]} {
+		b.WriteString(s)
+	}
```
mhutchinson, line 64: This feels like it would be easier to read as a regexp with capturing groups, which are then stitched together. But this works and it's at least hidden.

mhutchinson (approved):

serverless/internal/storage/fs/layout/layout.go
AlCutter: I thought about that (and about doing a `strings.Split` on `os.FilePathSeparator`, but figured it's quite tightly defined at the mo and this is likely cheaper.

As an aside, I was vaguely musing about having some sort of `geometry` file be part of the on-disk log with params for these sorts of paths, tile sizes, etc. Not sure it's currently worth the added complexity, though.

PR #286, opened 2021-04-01
Client sketch for map checks

mhutchinson:
The plan is to serve the map API primarily via a REST API. The server will use the sqlite DB for the first iteration, but the idea is that this could be swapped for static content similar to the serverless log.

This required a little rerouting to get the FWMeta back to where it could be checked.

### DIFF
--- a/binary_transparency/firmware/api/map.go
+++ b/binary_transparency/firmware/api/map.go
@@ -44,7 +44,7 @@ type MapCheckpoint struct {
 }
 
 // MapInclusionProof contains the value at the requested key and the proof to the
-// requested tree size.
+// requested Checkpoint.
 type MapInclusionProof struct {
 	Key   []byte
 	Value []byte
--- a/binary_transparency/firmware/api/map.go
+++ b/binary_transparency/firmware/api/map.go
@@ -14,6 +14,8 @@
 
 package api
 
+import "fmt"
+
 // AggregatedFirmware represents the results of aggregating a single piece of firmware
 // according to the rules described in #Aggregate().
 type AggregatedFirmware struct {
@@ -27,4 +29,29 @@ type AggregatedFirmware struct {
 type DeviceReleaseLog struct {
 	DeviceID  string
 	Revisions []uint64
+}
+
+// MapCheckpoint is a commitment to a map built from the FW Log at a given size.
+// The map checkpoint contains the checkpoint of the log this was built from, with
+// the number of entries consumed from that input log. This allows clients to check
+// they are seeing the same version of the log as the map was built from. This also
+// provides information to allow verifiers of the map to confirm correct construction.
+type MapCheckpoint struct {
+	LogCheckpoint LogCheckpoint
+	LogSize       uint64
+	RootHash      []byte
+	Revision      uint64
+}
+
+// MapInclusionProof contains the value at the requested key and the proof to the
+// requested tree size.
+type MapInclusionProof struct {
+	Key   []byte
+	Value []byte
+	Proof [][]byte
+}
+
+// String returns a compact printable representation of an InclusionProof.
+func (l MapInclusionProof) String() string {
+	return fmt.Sprintf("{key: 0x%x, value: 0x%x, proof: %x}", l.Key, l.Value, l.Proof)
 }
--- a/binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
+++ b/binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"reflect"
 
 	"github.com/golang/glog"
 	"github.com/google/trillian-examples/binary_transparency/firmware/api"
@@ -38,6 +39,7 @@ import (
 type FlashOpts struct {
 	DeviceID      string
 	LogURL        string
+	MapURL        string
 	WitnessURL    string
 	UpdateFile    string
 	Force         bool
@@ -53,7 +55,7 @@ func Main(opts FlashOpts) error {
 
 	up, err := readUpdateFile(opts.UpdateFile)
 	if err != nil {
-		return fmt.Errorf("Failed to read update package file: %w", err)
+		return fmt.Errorf("failed to read update package file: %w", err)
 	}
 	// TODO(al): check signature on checkpoints when they're added.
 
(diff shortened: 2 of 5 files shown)


binary_transparency/firmware/api/map.go
```diff
@@ -28,3 +30,28 @@ type DeviceReleaseLog struct {
...
+// provides information to allow verifiers of the map to confirm correct construction.
+type MapCheckpoint struct {
+	LogCheckpoint LogCheckpoint
+	LogSize       uint64
+	RootHash      []byte
+	Revision      uint64
+}
+
+// MapInclusionProof contains the value at the requested key and the proof to the
+// requested tree size.
```
AlCutter, line 47: To the requested tree size or to the map root?

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -107,10 +113,39 @@ func Main(opts FlashOpts) error {
...
 	}
 
-	// TODO(mhutchinson): query the map:
-	// 1. Check that the map root commits to a log root consistent with the checkpoints above
-	// 2. Look up the aggregation under the key: fmt.Sprintf("summary:%d", pb.InclusionProof.LeafIndex)
+	if len(opts.MapURL) > 0 {
+		mc := client.MapClient{}
+		mcp, err := mc.MapCheckpoint()
+		if err != nil {
+			return fmt.Errorf("failed to get map root: %w", err)
```
AlCutter: root -> checkpoint for consistency

binary_transparency/firmware/internal/verify/bundle.go
```diff
@@ -34,53 +34,53 @@ type ConsistencyProofFunc func(from, to uint64) ([][]byte, error)
...
 		}
 	}
-	return proofBundle, nil
+	return proofBundle, fwMeta, nil
 }
 
-// BundleValidateWitness verifies the received bundle against witness checkpoint
-func BundleValidateWitness(pb api.ProofBundle, wc api.LogCheckpoint, cpFunc ConsistencyProofFunc) error {
+// BundleValidateRemote verifies the received bundle against a remotely retrieved checkpoint (e.g. witness).
+func BundleValidateRemote(pb api.ProofBundle, rc api.LogCheckpoint, cpFunc ConsistencyProofFunc) error {
```
AlCutter, line 63: The naming in here was intended to be read with the package name (e.g. "verify Bundle For Update"), how about renaming this method to `verify.Consistency` and just pass in `pb.Checkpoint` and `rc`?

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -157,36 +164,77 @@ func getConsistencyFunc(c *client.ReadonlyClient) func(from, to uint64) ([][]byt
...
+		return fmt.Errorf("witness_url is invalid: %w", err)
+	}
+	wc := client.WitnessClient{URL: wURL}
+
 	wcp, err := wc.GetWitnessCheckpoint()
 	if err != nil {
 		return fmt.Errorf("failed to fetch the witness checkpoint: %w", err)
 	}
+	if err := verifyRemote(c, pb, *wcp); err != nil {
+		return fmt.Errorf("failed to verify update with witness: %w", err)
```
AlCutter, line 198: "verify update consistency against" just to be clear that we're not sending the update off to the witness or anything

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -157,36 +164,77 @@ func getConsistencyFunc(c *client.ReadonlyClient) func(from, to uint64) ([][]byt
...
+	}
+	return nil
+}
+
+func verifyMap(c *client.ReadonlyClient, pb api.ProofBundle, fwMeta api.FirmwareMetadata, mapURL string) error {
```
AlCutter, line 203: `verifyAggregation` (or similar)?  - all the checkpoints/consistency, etc. are leading up to that "is it known bad?" at the end which is answered by the aggregation

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -157,36 +164,77 @@ func getConsistencyFunc(c *client.ReadonlyClient) func(from, to uint64) ([][]byt
...
+	if !afw.Good {
+		return errors.New("firmware is marked as bad")
+	}
+	return nil
+}
 
-	if wcp.TreeSize == 0 {
-		return fmt.Errorf("No witness checkpoint to verify")
+// verifyRemote checks that an update package is consistent with remotely fetched Checkpoint.
+func verifyRemote(c *client.ReadonlyClient, pb api.ProofBundle, rcp api.LogCheckpoint) error {
```
AlCutter, line 232: is this really `verifyConsistency` (same as the `verify` package)(?)

AlCutter (commented):
Few thoughts

binary_transparency/firmware/api/map.go
mhutchinson: Checkpoint, good catch.

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: Done.

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: `verifyAnnotations` as this will have more (if more is added).

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: Inlined this method out of existence.

binary_transparency/firmware/internal/verify/bundle.go
mhutchinson: renamed to `verify.BundleConsistency` as it verifies the bundle is consistent. It still needs a little other spice than the bundle and a checkpoint as its otherwise stateless and needs access to a client. That's what the `cpFunc` provides.

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
AlCutter: Ah, yes - that's what I _meant_ :)

AlCutter (approved):

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: Also done now.

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: :-)

PR #290, opened 2021-04-01
[serverless] Use leafhash files as hash->seq index

AlCutter:
This change allows clients who only hold the leaf primage to figure out the sequence number for that leaf and build the corresponding inclusion proof.


serverless/internal/storage/fs/fs.go
```diff
@@ -157,8 +158,8 @@ func (fs *Storage) Sequence(leafhash []byte, leaf []byte) (uint64, error) {
 	// If there is one, it's a symlink to the sequence file, so read that back
```
mhutchinson, line 158: This still references symlinks.

mhutchinson (approved):
Can you add 1-2 lines of motivation ("why" this CL, not "you can do it, Al!") to the PR description?

serverless/internal/storage/fs/fs.go
AlCutter: ta, done

PR #293, opened 2021-04-12
Updated to latest version of trillian

mhutchinson:
All the changes are fallout from either direct changes to trillian API, or because the update to trillian updated the proto API, and that now has a bunch of stuff which is deprecated and thus blocking due to our usage of the linter.

This fixes the nil signature problem seen because we're running FT against a non-pinned version of Trillian images. The fix is very indirect; it's the trillian client that removes this check.


binary_transparency/firmware/cmd/ft_personality/impl/ft_personality.go
```diff
@@ -32,8 +32,8 @@ import (
 	"github.com/google/trillian-examples/binary_transparency/firmware/cmd/ft_personality/internal/trillian"
 	"github.com/gorilla/mux"
 
-	_ "github.com/google/trillian/merkle/rfc6962" // Load hashers
-	_ "github.com/mattn/go-sqlite3"               // Load drivers for sqlite3
+	_ "github.com/google/trillian/merkle/rfc6962/hasher" // Load hashers
```
AlCutter, line 35: This package doesn't have the side-effect registration of the hasher any more, so I think it's safe to remove this line (but we can do this in a follow-up).

AlCutter (approved):

PR #295, opened 2021-04-12
[serverless] Add some initial storage layout docs

AlCutter:


serverless/internal/layout/README.md
```diff
@@ -0,0 +1,86 @@
+Layout
+======
+
+This doc describes the on-disk layout and format of the serverless log files.
+
+> :warning: this is still under development, and the exact layout and contents
+> of the serverless log structure may well change.
+
+Broadly, the layout consists of a self-contained directory hierarchy whose
+contents represents the entire state of the log. With one exception, the
```
mhutchinson, line 10: nit: my first reading of this had "With one exception" following on from the previous statement about self-contained directory. It isn't that. But it could just be me.

serverless/internal/layout/README.md
```diff
@@ -0,0 +1,86 @@
...
+contents and structure of this directory is designed to allow it to be safely
+and indefinitely cached by clients.
+
+Inside the directory you'll find:
+
+ * :file_folder: leaves/
+ * :file_folder: seq/
+ * :file_folder: tile/
+ * :page_facing_up: state
```
mhutchinson, line 19: nit: the order of these not matching the order below is a little distracting.

serverless/internal/layout/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+tile/
+-----
+`tile/` contains the internal nodes of the log tree.
+
+The nodes are grouped into "tiles" - these are 8 level deep "sub trees".
+Each tile exists at a particular location in tile-space, identified by its
+`layer` and `index`.
+
+A *perfect* tree of size 65536 would be represented by two layers of tiles:
```
mhutchinson, line 59:
```suggestion
A *perfect* tree of size 2^16 would be represented by two layers of tiles:
```
That's just way easier for me. But could just be me.

serverless/internal/layout/README.md
```diff
@@ -0,0 +1,86 @@
...
+The contents of the file is simply the hex ASCII string representation of the
+index.
+
+tile/
+-----
+`tile/` contains the internal nodes of the log tree.
+
+The nodes are grouped into "tiles" - these are 8 level deep "sub trees".
+Each tile exists at a particular location in tile-space, identified by its
+`layer` and `index`.
```
mhutchinson, line 57: Consider strata instead of layer. It would be consistent with the map terminology.

serverless/internal/layout/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+A *perfect* tree of size 65536 would be represented by two layers of tiles:
+ - layer 0, 256 tiles wide, containing all the leaf hashes of the log along with
+   the 7 bottom-most layers of internal nodes.
+ - layer 1, 1 tile wide, containing the next 8 layers of internal nodes.
+
+Given a `layer` of `0x0123456789` and an `index` of `0xef`, the corresponding
+fully populated tile file would be found at `.../tile/ef/0123/45/67/89`.
```
mhutchinson, line 65: This seems surprising that the index goes before the layer. Might be useful to explain why?

mhutchinson (approved):

serverless/internal/layout/README.md
AlCutter: Flipped around, does that scan better for you?

serverless/internal/layout/README.md
AlCutter: rejiggered

serverless/internal/layout/README.md
AlCutter: sure, done :)

serverless/internal/layout/README.md
AlCutter: changed

serverless/internal/layout/README.md
AlCutter: or... fix it;  it was Just Plain Wrong(tm), and is indeed stratum followed by index :) good spot, ta!

PR #303, opened 2021-04-16
Implemented map server methods

mhutchinson:
The map server now supports:
 * lookup of tiles by the path to the root of the tile
 * getting an aggregation for FW logged at a given index

Updated the client pseudocode to better match the shape of the server API.

This is based on top of #302.

### DIFF
--- a/binary_transparency/firmware/cmd/ftmapserver/impl/ftmapserver.go
+++ b/binary_transparency/firmware/cmd/ftmapserver/impl/ftmapserver.go
@@ -17,10 +17,12 @@ package impl
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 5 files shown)


binary_transparency/firmware/cmd/ftmapserver/impl/ftmapserver.go
```diff
@@ -119,7 +124,93 @@ func (s *Server) getCheckpoint(w http.ResponseWriter, r *http.Request) {
...
+func parseBase64Param(r *http.Request, name string) ([]byte, error) {
+	v := mux.Vars(r)
+	b, err := base64.URLEncoding.DecodeString(v[name])
+	if err != nil {
+		return nil, fmt.Errorf("%s should be URL-safe base64 (%q)", name, err)
+	}
+	return b, nil
+}
+
+func parseIntParam(r *http.Request, name string) (uint64, error) {
```
AlCutter, line 209: nit: consider calling this `parseUintParam`?

AlCutter (approved):

PR #304, opened 2021-04-20
Flash tool uses aggregations from map to check for malware

mhutchinson:
The map checkpoint and tiles are now fetched from the server. `mapclient` generates a compact inclusion proof from this data, which is then verified by `flash_tool`.

This architecture is belt and braces, but my thinking evolved during this PR. Initially the plan was to have a very simple server which could work like the "serverless" log and just serve static content. However, as I started to write the map client, I realized that this pushed a lot of work onto the clients, which might have limited compute power. I've thus written the logic in the `mapclient` as effectively untrusted, which means that this work could be done on a server, meaning much less hashing and space requirements for the verifying client.

Raising this as a draft PR to get an early review on this approach, before I write the tests that are missing.

### DIFF
--- a/binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
+++ b/binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
@@ -242,7 +242,7 @@ func verifyAnnotations(ctx context.Context, c *client.ReadonlyClient, pb api.Pro
 	// and everything else. In a production system, it is likely that the proof generation
 	// would live elsewhere (e.g. in the OTA packaging process), and the shorter proof
 	// bundle would be provided to the device.
-	afw, ip, err := mc.Aggregation(ctx, mcp.Revision, pb.InclusionProof.LeafIndex)
+	preimage, ip, err := mc.Aggregation(ctx, mcp.Revision, pb.InclusionProof.LeafIndex)
 	if err != nil {
 		return fmt.Errorf("failed to get map value for %q: %w", pb.InclusionProof.LeafIndex, err)
 	}
@@ -252,10 +252,6 @@ func verifyAnnotations(ctx context.Context, c *client.ReadonlyClient, pb api.Pro
 		return fmt.Errorf("received inclusion proof for key %x but wanted %x", ip.Key, kbs[:])
 	}
 	// 2. Check: the proof is for the correct value.
-	preimage, err := json.Marshal(afw)
-	if err != nil {
-		return fmt.Errorf("failed to marshal aggregation to compute preimage: %w", err)
-	}
 	leafID := tree.NewNodeID2(string(kbs[:]), 256)
 	hasher := coniks.Default
 	expectedCommitment := hasher.HashLeaf(api.MapTreeID, leafID, preimage)
@@ -286,8 +282,12 @@ func verifyAnnotations(ctx context.Context, c *client.ReadonlyClient, pb api.Pro
 		return fmt.Errorf("inclusion proof calculated root %x but wanted %x", calc, mcp.RootHash)
 	}
 
+	var agg api.AggregatedFirmware
+	if err := json.NewDecoder(bytes.NewReader(preimage)).Decode(&agg); err != nil {
+		return fmt.Errorf("failed to decode aggregation: %w", err)
+	}
 	// Now we're certain that the aggregation is contained in the map, we can use the value.
-	if !afw.Good {
+	if !agg.Good {
 		return errors.New("firmware is marked as bad")
 	}
 	return nil
--- a/binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
+++ b/binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
@@ -18,6 +18,8 @@
 package impl
 
 import (
+	"bytes"
+	"context"
 	"crypto/sha512"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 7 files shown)


AlCutter:
> However, as I started to write the map client, I realized that this pushed a lot of work onto the clients, which might have limited compute power.

Right - there are definitely 3 distinct operations/roles:
- storing/serving nodes/tiles
- proof building
- proof verification

the middle one can be (and historically has always been) bundled into either the server (e.g. CT) or the verifier (e.g golang), but there are definitely cases where it makes sense to keep it separate, e.g. here in FT I suspect that the OTA bundle builder/server would be fetching (and "reconstituting") the map tiles and deriving concise proofs to push down to the `flash_tool` as part of the OTA bundle.

mhutchinson:
I've tidied this up and it's ready for review. I agree with your clear writeup about these 3 roles in the proof story. At the moment the proof construction is on the client side, but it's written separately enough that this could easily be moved elsewhere. This is a feature :-)

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -224,13 +236,65 @@ func verifyAnnotations(c *client.ReadonlyClient, pb api.ProofBundle, fwMeta api.
...
 	if err != nil {
 		return fmt.Errorf("failed to get map value for %q: %w", pb.InclusionProof.LeafIndex, err)
 	}
+	// 1. Check: the proof is for the correct key.
+	kbs := sha512.Sum512_256([]byte(fmt.Sprintf("summary:%d", pb.InclusionProof.LeafIndex)))
+	if !bytes.Equal(ip.Key, kbs[:]) {
+		return fmt.Errorf("received inclusion proof for key %x but wanted %x", ip.Key, kbs[:])
+	}
+	// 2. Check: the proof is for the correct value.
+	preimage, err := json.Marshal(afw)
```
AlCutter, line 255: This might be brittle, maybe `mc.Aggregation` should return the raw leaf bytes and leave the parsing to the caller?

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -224,13 +236,65 @@ func verifyAnnotations(c *client.ReadonlyClient, pb api.ProofBundle, fwMeta api.
...
+		}
+		left, right := calc, sib
+		if !isLeftChild(stem) {
+			left, right = right, left
+		}
+		calc = hasher.HashChildren(left, right)
+	}
+	if !bytes.Equal(calc, mcp.RootHash) {
+		return fmt.Errorf("inclusion proof calculated root %x but wanted %x", calc, mcp.RootHash)
+	}
```
AlCutter, line 287: Worth pulling this out into a separate func to keep things small?

mhutchinson: That was one of the approaches I tried on the journey, but the number of parameters ends up being similar to the number of lines extracted! This needs:
 * The expected leaf commitment (`calc`)
 * `hasher`
 * `ip`
 * `leafID`
 * `mcp.RootHash`

If you still think it's worth it then I'll do it, but it felt messy to me.

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: There are comments in to this effect, but I will just resolve them now I guess. I went with this slightly risque approach because it led to a neat implementation of `fetch` inside the client that avoided needing to mess about with io closers when reading the body.

mhutchinson: This is addressed

AlCutter: Yeah, I spotted them a bit further down, but I figured this was at least worth a discussion, and maybe I'd try to convince you to change it :)

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
AlCutter: Your call, don't feel super strongly about it, and if this verification needs to be done by something else at some point in the future I guess it can always be extracted then.

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
```diff
@@ -286,8 +282,12 @@ func verifyAnnotations(ctx context.Context, c *client.ReadonlyClient, pb api.Pro
 		return fmt.Errorf("inclusion proof calculated root %x but wanted %x", calc, mcp.RootHash)
 	}
 
+	var agg api.AggregatedFirmware
+	if err := json.NewDecoder(bytes.NewReader(preimage)).Decode(&agg); err != nil {
```
AlCutter, line 286: Can do ` ... := json.Unmarshal(preimage, &agg); ...`

binary_transparency/firmware/internal/client/mapclient.go
```diff
@@ -88,22 +87,27 @@ func (c *MapClient) Aggregation(ctx context.Context, rev uint64, fwIndex uint64)
 		errs.Go(func() error {
 			path := kbs[:i]
 			var t api.MapTile
-			err := c.fetch(fmt.Sprintf("%s/in-revision/%d/at-path/%s", api.MapHTTPGetTile, rev, base64.URLEncoding.EncodeToString(path)), &t)
+			tbs, err := c.fetch(fmt.Sprintf("%s/in-revision/%d/at-path/%s", api.MapHTTPGetTile, rev, base64.URLEncoding.EncodeToString(path)))
 			if err != nil {
 				return err
 			}
+			if err := json.NewDecoder(bytes.NewReader(tbs)).Decode(&t); err != nil {
```
AlCutter, line 94: Same here

AlCutter (approved):
small suggestions inline

binary_transparency/firmware/cmd/flash_tool/impl/flash_tool.go
mhutchinson: I looked for this under json.Decode and then wrote it off as a figment of my imagination that such an obvious thing would be possible. Thanks for the pointer!

PR #309, opened 2021-04-28
[serverless] Use simple format for serialised checkpoints

AlCutter:


serverless/api/state.go
```diff
@@ -15,6 +15,17 @@
...
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// CheckpointHeader is the first line of a marshaled log checkpoint.
+const CheckpointHeader = "Log Checkpoint v0"
```
mhutchinson, line 27: Stick v0 in the const name (or make it private/lowercase)?

serverless/api/state.go
```diff
@@ -24,6 +35,44 @@ type LogState struct {
...
+		return errors.New("invalid checkpoint - incorrect header")
+	}
+	size, err := strconv.ParseUint(l[1], 10, 64)
+	if err != nil {
+		return fmt.Errorf("invalid checkpoint - size invalid: %w", err)
+	}
+	rh, err := base64.StdEncoding.DecodeString(l[2])
+	if err != nil {
+		return fmt.Errorf("invalid checkpoint - invalid roothash: %w", err)
+
```
mhutchinson, line 70:
```suggestion
```

serverless/api/state.go
```diff
@@ -24,6 +35,44 @@ type LogState struct {
...
+	}
+	s.Size, s.RootHash = size, rh
```
mhutchinson, line 72: Consider checking the size of `rh`.  The doc for `RootHash` says it's 256 bits so may as well be paranoid on inputs. Though I realize this totally makes your tests less fun. So YMMV :-)

mhutchinson (approved):
nit: also change PR description to s/add a/changed to/ ; this changes the format used.

serverless/api/state.go
AlCutter: done

serverless/api/state.go
AlCutter: done

serverless/api/state.go
AlCutter: I think that's probably the layer above (unless I make the struct have a `[32]byte` for `RootHash`?), maybe I'll update the comment instead :)

AlCutter (commented):
> nit: also change PR description to s/add a/changed to/ ; this changes the format used.

Good call, done

AlCutter:
(rebased)

PR #310, opened 2021-04-28
[serverless] Store tiles in a slightly leaner format than JSON

AlCutter:
This PR implements a simple custom text format for serialising the log tiles, rather than directly using JSON.
The format is trying to find a sweet spot between a full binary representation and something which is easy to read/debug but without too much extra clutter.

An example tile in this format looks like this:
```
32
4
qsZs16ec5AEtgHYv6O7Dp38i0cpBRcP0zuAi5+/NWZ0=
z9+lOZmJFyF/Mpcpq483zbJbCGFye9Qq0qH5xQ+3UyM=
ycWD0Xb0jgtbC1nQ6Of7QWpFxlASqs3nQC0vmQ8mxcw=
qiSZ0B/Hha39/LXy6yETE+CZxGE0SDtDM2IjhxLOGAo=
hd9aDSQrndkekVXXrt2weXcStDte0WrHMwKwHZCQLo4=
XnuygG2bXSR9CnvgLx9AlExUsMD1pm9ouR2l1KwsQE4=
Xa/RR4kVQaZZiL5oa3epz0H4dgtdELmfCd3bpTyZVnA=
```

### DIFF
--- a/serverless/api/state.go
+++ b/serverless/api/state.go
@@ -15,6 +15,14 @@
 // Package api contains the "public" API/artifacts of the serverless log.
 package api
 
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
 // LogState represents the state of a serverless log
 type LogState struct {
 	// Size is the number of leaves in the log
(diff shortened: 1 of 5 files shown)


mhutchinson (approved):
LGTM, but please update PR description to make it clearer that this moves away from a standard JSON format in tile representation to a custom serialization format. The PR description sounds like its a small blackbox hidden "storage" change, but the representation of these tiles is the API.

PR #316, opened 2021-05-06
[serverless] Ensure sequence only runs when there are entries to add 

paulmattei:


serverless/cmd/sequence/main.go
```diff
@@ -39,13 +39,27 @@ var (
 
 func main() {
 	flag.Parse()
+	var err error
```
AlCutter, line 42: No need to move this line - it's used below by the `if *create / else` stuff - it's to have the `err` _outside_ the scope of the if so we just have a single `if err != nil` check which covers both cases (on L75).

Here, though, you're declare-and-assign'ing an `err` which exists only in the scope of the if (`if err := ...`), consequently we don't need the var declaration.

(This is why you're getting the warning from the `LGTM` check on this PR, BTW.)

serverless/cmd/sequence/main.go
```diff
@@ -39,13 +39,27 @@ var (
...
+
+	toAdd, err := filepath.Glob(*entries)
+	if err != nil {
+		glog.Exitf("Failed to glob entries %q: %q", *entries, err)
+	}
+	if toAdd == nil {
```
AlCutter, line 48: If we're dealing with slices, then it's better form to always use `len(thing)` for these checks - it'll happily return `0` for zero sized slices, as well as `nil` slices.

serverless/cmd/sequence/main.go
```diff
@@ -39,13 +39,27 @@ var (
...
+		glog.Exitf("Sequence must be run with at least one valid entry")
```
AlCutter, line 49: Can use `glog.Exit` here (no `f`) since this is just static text (no format specifiers/params)

serverless/cmd/sequence/main.go
```diff
@@ -39,13 +39,27 @@ var (
...
 	// init storage
 	var st *fs.Storage
-	var err error
 	if *create {
 		st, err = fs.Create(*storageDir, h.EmptyRoot())
+		r, _ := api.LogState{}.MarshalText()
+		if err := st.WriteLogState(r); err != nil {
+			glog.Exitf("Failed to create empty log state: %q", err)
+		}
+
```
AlCutter, line 62: Don't need this bit any more

AlCutter (approved):

PR #322, opened 2021-05-12
Persist the monitor state to avoid repeated work

mhutchinson:
Without this, the monitor starts from the beginning of the log on each run. This is wasted effort in the best-case, but can actually cause a lot of spam annotations to the log in the worst case.

The state file can always be removed locally if the monitor code is updated and needs to process the whole log again.

### DIFF
--- a/binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
+++ b/binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
@@ -28,6 +28,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"os"
 	"regexp"
 	"time"
 
@@ -48,11 +49,15 @@ type MonitorOpts struct {
 	Keyword      string
 	Matched      MatchFunc
 	Annotate     bool
+	StateFile    string
 }
 
 func Main(ctx context.Context, opts MonitorOpts) error {
 	if len(opts.LogURL) == 0 {
 		return errors.New("log URL is required")
+	}
+	if len(opts.StateFile) == 0 {
+		return errors.New("state file is required")
 	}
 
 	ftURL, err := url.Parse(opts.LogURL)
@@ -65,15 +70,23 @@ func Main(ctx context.Context, opts MonitorOpts) error {
 
 	c := client.ReadonlyClient{LogURL: ftURL}
 
-	// TODO(mhutchinson): This checkpoint and tracker for number of processed entries
-	// should be serialized so the monitor persists its golden state between runs.
-	// In particular this will prevent the annotator from adding multiple annotations
-	// for the same firmware each time the monitor is started.
+	// Initialize the checkpoint from persisted state.
 	var latestCP api.LogCheckpoint
-	var head uint64
+	if state, err := os.ReadFile(opts.StateFile); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to read state: %w", err)
+		}
+		// Silently continue if the file doesn't exist.
+		// This could fail here unless a force flag is provided, for better security.
+	} else {
+		if err := json.Unmarshal(state, &latestCP); err != nil {
+			return fmt.Errorf("failed to read state: %w", err)
+		}
+	}
+	head := latestCP.TreeSize
 	follow := client.NewLogFollower(c)
 
-	glog.Infof("Monitoring FT log %q...", opts.LogURL)
+	glog.Infof("Monitoring FT log (%q) starting from index %d", opts.LogURL, head)
 	cpc, cperrc := follow.Checkpoints(ctx, opts.PollInterval, latestCP)
 	ec, eerrc := follow.Entries(ctx, cpc, head)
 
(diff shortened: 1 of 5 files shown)


mhutchinson:
Sorry for the noise. This is ready for review now.

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
```diff
@@ -48,12 +50,16 @@ type MonitorOpts struct {
...
 	Matched      MatchFunc
 	Annotate     bool
+	StateFile    string
 }
 
 func Main(ctx context.Context, opts MonitorOpts) error {
 	if len(opts.LogURL) == 0 {
 		return errors.New("log URL is required")
 	}
+	if len(opts.StateFile) == 0 {
```
yogeshbdeshpande, line 60: Is this check not contradictory to check on Line 76? If the StateFile is not provided, perhaps we should continue, as we have a mechanism to continue, in absence of a StateFile ?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
```diff
@@ -65,15 +71,23 @@ func Main(ctx context.Context, opts MonitorOpts) error {
...
-	// In particular this will prevent the annotator from adding multiple annotations
-	// for the same firmware each time the monitor is started.
+	// Initialize the checkpoint from persisted state.
 	var latestCP api.LogCheckpoint
-	var head uint64
+	if state, err := ioutil.ReadFile(opts.StateFile); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("failed to read state: %w", err)
+		}
+		// Silently continue if the file doesn't exist.
```
yogeshbdeshpande, line 80: Perhaps we should generate a Warning here and then progress?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
AlCutter: I think it would be unusual to want to operator a monitor without a state _filename_ configured - it'd be vulnerable to split views during any point at which the monitor was not running.

The section around L77 deals with the edge case of the first time the monitor is run (post this PR) and the file doesn't _yet_ exist (but will shortly be created).

AlCutter (approved):

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: Good call; done.

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: I initially coded this as optional, but then came down hard on the opinion that Al has laid out; any monitor worth its salt must absolutely be persisting its golden checkpoint.

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
```diff
@@ -89,60 +103,77 @@ func Main(ctx context.Context, opts MonitorOpts) error {
...
-			sc := client.SubmitClient{
-				ReadonlyClient: &c,
-			}
-			if err := sc.PublishAnnotationMalware(js); err != nil {
-				return fmt.Errorf("failed to publish annotation: %w", err)
-			}
+		glog.V(1).Infof("Annotating %s", ms)
+		js, err := createStatementJSON(ms)
+		if err != nil {
+			glog.Warningf("failed to create annotation: %q", err)
```
yogeshbdeshpande, line 168: The earlier Error points in the old code on these have been turned now into Warnings.
Should we not fail, if the statement creation fails?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
```diff
@@ -89,60 +103,77 @@ func Main(ctx context.Context, opts MonitorOpts) error {
...
+			return
+		}
+		sc := client.SubmitClient{
+			ReadonlyClient: &c,
+		}
+		if err := sc.PublishAnnotationMalware(js); err != nil {
+			glog.Warningf("failed to publish annotation: %q", err)
```
yogeshbdeshpande, line 175: Same comment as 168!

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
```diff
@@ -89,60 +103,77 @@ func Main(ctx context.Context, opts MonitorOpts) error {
...
+				return fmt.Errorf("failed to marshal checkpoint: %w", err)
+			}
+			ioutil.WriteFile(opts.StateFile, bs, 0o755)
+			glog.Infof("Persisted state: %v", entry.Root)
 		}
+	}
+}
 
-		glog.Infof("Found firmware (@%d): %s", entry.Index, meta)
+func processEntry(entry client.LogEntry, c client.ReadonlyClient, opts MonitorOpts, matcher *regexp.Regexp) {
```
yogeshbdeshpande, line 121: is it not better to return an Error to the caller if processEntry fails?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: Good spot. I've made these checks consistent with the other checks in the monitor, which were all warnings. The history of this class is that initially everything was a warning, and then I inadvertently introduced the dissonance when I added the annotation code and made things throw errors. The refactoring in this PR made that dissonance clear, so I've resolved it by making the annotation code consistent with everything else.

If we want to, we could follow up with a new PR to make all of these warnings be errors. I'm not sure about that though. For demoware, this strikes a reasonable middle ground.

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: Sure, we could take that option. I was going to do that but then lines {131, 140, 146} all looked really suspicious. As that was the older code than the lines in here which already returned errors, I gave them the right of way.

Are you saying you'd prefer all return paths in here that currently warn to return an error?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
yogeshbdeshpande: I will be OK, if we track it with a new issue, to classify which of these warnings are genuine errors which we should block monitor from further operation, versus which we can continue to proceed as Warnings. This way we can revisit this section of code and then proceed with this PR in the state as it is, marking the comment to resolved!

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
yogeshbdeshpande: For now we can leave this comment tagged with the new issue then and revisit again!
Perhaps, when we revisit this function to classify errors, then we will have anyway need to change the template of this function to return an error. For now, you can continue the style as now we agreed to treat all of them as Warnings, with an issue open!

yogeshbdeshpande (approved):

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: I've changed the code to return errors from all places, and the calling code then logs the warning. There is a TODO there to consider a strict processing mode which will exit instead of logging the warning. I think this is a reasonable tidy compromise?

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
mhutchinson: This is the same conversation as above, so resolving this and we'll track up there :-)

binary_transparency/firmware/cmd/ft_monitor/impl/ft_monitor.go
yogeshbdeshpande: Thanks, Yes much better now!

PR #328, opened 2021-05-17
Added link from main README to verifiable maps

mhutchinson:


yogeshbdeshpande:
I will take some time to review this in detail, but a first look, wanted to check with you, does this update also needs a paired update to Claimant Model? I recollect in the strawman:
go/claimant-model-maps needs to be written up externally at some point. It builds on the Claimant Model and introduces new terms

binary_transparency/firmware/README.md
```diff
@@ -340,14 +340,16 @@ go run cmd/publisher/publish.go --logtostderr --v=2 --timestamp="2020-10-10T23:0
 > Anybody else running a monitor also knows that malicious firmware has been
 > logged and can raise the alarm.
 
-Developer Notes
----------------
 
-This section is not intended for a general audience.
-The intended audience is developers of FT that have a deployment and need to inspect or modify it.
+Further Work: Annotations and Verifiable Summaries
```
yogeshbdeshpande, line 344: Perhaps, the new feature, needs to be well integrated/gelled with the existing story as a continuity.
--
As soon as the Monitor has seen the malicious Firmware, now the Monitor can immediately Annotate this firmware as “BAD” and submit this annotation to the log as a new entry.

<< Need to state here who will run the Map Server to provide the extra layer, introduced by the Map>>



Any device getting this malicious firmware in the update from publisher, has now the required means to detect it been bad. It will use the Map URL to query the added feature of “Verifiable Map” to fetch all Annotations about this firmware to conclude it been BAD. This feature thus prevents even a Hacked Firmware submitted to a Log as a Valid Entry been applied on the device!

mhutchinson: I very intentionally tried to make this a separate section. I think we should also do the same with the witness. In my view the main story in this readme should set up the basic flow and understanding, and then there are a number of additional optional quests you can also go in, in any order. Trying to make one uber-flow with everything in it would be intimidating, IMO. I think the writeup in this README should contain a brief overview of each of the optional extras, but I think we should leave a detailed discussion to the individual READMEs.

It's a good observation that we could write up the Maps CM mapping somewhere. I don't think it's strictly required right now, but I'll get it done.

mhutchinson: Whoa, I forgot that Past Martin already did write up the CM for Maps: https://github.com/google/trillian/blob/master/docs/claimantmodel/Maps.md

yogeshbdeshpande: Thanks, I agree, lets discuss this (possibly in todays meeting) what we should write in the top level README.md

Also I realized the point that Witness also needs some sort of documentation here, while reviewing your change. Thanks anyway for pointing again, will do that as well :)

Agree your point on CM, once we agree on how we document the optional features, we can break down the CM components specific to these  optional entities in a similar fashion.

yogeshbdeshpande (approved):

AlCutter (approved):

PR #336, opened 2021-05-20
Add link to Optional STH Witness feature

yogeshbdeshpande:


binary_transparency/firmware/docs/design/README.md
```diff
@@ -234,7 +234,8 @@ _discoverability_ that Firmware Transparency leverages.
 
 **STH Witness**:
 
-_Not yet implemented here._
+May use the [`ft_witness`](/binary_transparency/firmware/cmd/ft_witness) to
+support verfication of log operator, for consistent log operation:
```
AlCutter, line 238: how about `... to enable detection of log split views`?

wdyt?

yogeshbdeshpande: Yes, I was also planning to add something similar, however we have not explained the terms "split views and split view attacks till now, so was unsure. They have been explained in the STH Witness Page..

yogeshbdeshpande: Perhaps I can add:
to support additional verification of log operator by enabling detection of log split views.

yogeshbdeshpande: Please check this revision, with some corrections, on lines similar to what you had suggested!

AlCutter (approved):

PR #340, opened 2021-05-24
[Helloworld] Add signature checking to helloworld

AlCutter:


helloworld/client.go
```diff
@@ -17,54 +17,51 @@ package helloworld
...
+	// It is the job of the client to verify this inclusion proof in order to convince
 	// itself that the entry really has been included in the log.
-	ProveIncl(context.Context, *log.Checkpoint, []byte) (*trillian.Proof, error)
+	ProveIncl(context.Context, uint64, []byte) (*trillian.Proof, error)
 
 	// UpdateChkpt provides a new checkpoint and proves it is consistent
 	// with an old one.  Again, it is the job of the client to verify this
 	// consistency proof.
-	UpdateChkpt(context.Context, *log.Checkpoint) (*log.Checkpoint, *trillian.Proof, error)
+	UpdateChkpt(context.Context, uint64) ([]byte, *trillian.Proof, error)
```
mhutchinson, line 45: The comment doesn't match the signature.

helloworld/client.go
```diff
@@ -17,54 +17,51 @@ package helloworld
...
+	"golang.org/x/mod/sumdb/note"
 )
 
 // Personality is the front-end for the Trillian log.
 type Personality interface {
 	// Append stores an entry in the log and, once it is there, returns a
 	// new checkpoint that commits to the new entry (in addition to all
 	// previous ones).
-	Append(context.Context, []byte) (*log.Checkpoint, error)
+	Append(context.Context, []byte) ([]byte, error)
```
mhutchinson, line 35: I would like to see an alias for `[]byte` here to some new type like `RawCheckpoint` (it's not obvious at a first glance at this new sig that the input bytes are leaf preimage, and the output bytes are a raw checkpoint). We could later do the same with `RawProof`. I think it makes it easier to read and avoids type errors. A side bonus is potential for custom methods on there, but even if yagni, the other reasons motivate me.

mhutchinson (approved):

helloworld/client.go
AlCutter: Not sure what specific change you're asking for here - it still returns a checkpoint (albeit flattened), and a proof for an old checkpoint? But I suspect that part of the problem here is that the params on the interface methods don't have names, I'll fix that.

helloworld/client.go
AlCutter: Sure, can give it a go - but do you actually mean a type alias, or did you mean a named type definition of a `[]byte`?

In practice I'm not sure it'll actually help all that much with type errors (e.g. `note.Open` still takes a `[]byte`, and on-disk and over the wire representations are `[]byte`, etc.), and personally I still find redefinitions of primitive types to be a bit misleading ("yay, I'll just grab the sigs from this struct"... "huh?" ... "oh, it's a `[]byte` :(") but maybe I'm a dinosaur 🦖  and I guess if nothing else it's a bit more documentation.

I've gone with type definition for now, PTAL and see what you think.

helloworld/client.go
mhutchinson: It seems different this afternoon, but this morning "proves it is consistent with an old one" didn't seem appropriate when what was passed in was a `uint64`. Having names here helps.

mhutchinson (commented):
github has got itself wedged with comments and I can't continue the thread above. But what I typed (and couldn't then post) was: "Yeah I meant type definition. Thanks for teaching me the nuance here. This reads a little clearer to me, and I suspect the average punter that this example is targeted at. But I may live to eat my words :-)"

AlCutter:
> github has got itself wedged with comments and I can't continue the thread above. But what I typed (and couldn't then post) was: "Yeah I meant type definition. Thanks for teaching me the nuance here. This reads a little clearer to me, and I suspect the average punter that this example is targeted at. But I may live to eat my words :-)"

Yeah - that happened to me the other day too!

Cool, thanks - I'll send questions your way if they come :)

PR #342, opened 2021-05-25
[Format] Add signed checkpoint "documentation" type

AlCutter:
I thought I'd try it out across the repo and see how it went - wdyt?


mhutchinson (commented):
I'm on the fence with this one. I proposed this happily for `helloworld` because this is a beginner's introduction to these concepts. As for the whole API... I'm 100% ambivalent. And I think that means the reasonable change is no change.

If we did want to do this, I would consider renaming `SignedCheckpoint` to `CheckpointStatement`.

PR #344, opened 2021-05-26
Update serverless action to use latest

AlCutter:


smeiklej (approved):
LGTM (modulo being frightened).

PR #345, opened 2021-05-26
[serverless] More go get/tidy on action

AlCutter:


smeiklej (approved):
LGTM with the same caveat (terror) as before.

PR #358, opened 2021-06-18
[serverless] Expose client support

AlCutter:


serverless/client/client.go
```diff
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package client is a client for the serverless log.
+// Package client provides client support for interacting with a serverless
+// log.
+//
+// See the /cmd/client package in this repo for an example of using this.
```
mhutchinson, line 18:
```suggestion
// See the /serverless/cmd/client package in this repo for an example of using this.
```

?

mhutchinson (approved):
Minor concern that `client.Fetcher` isn't exactly descriptive, but without seeing some more representative usages in calling code I have no proposals for anything better.

serverless/client/client.go
AlCutter: The `repo` in the comment is trillian-examples, so I just fixed the path.

PR #359, opened 2021-06-22
Adding ethereum witness contracts

sinasabet81:
This PR adds the contracts for the Ethereum Witness. The files of interest are the Solidity contracts, under `contracts/`, and their Python tests under `tests/`.

See the README for info on how to run the tests locally.


witness/ethereum/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+## To run tests using docker
+
+- cd into this directory
+- `docker run --rm -it $(docker build -q .)`
+  - This builds the container, runs it, and removes it after the run is
+    complete.
+
+## Explaination
```
smeiklej, line 44: "Explanation"

witness/ethereum/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+The Ethereum Witness contracts act as an on-chain registry of merkle tree
```
smeiklej, line 46: Should be Merkle (not merkle).

witness/ethereum/README.md
```diff
@@ -0,0 +1,52 @@
...
+log states that are otherwise maintained off-chain. The contract handles
+updates to logs and helps ensure that log operators cannot provide a split
+view of the log prior to a confirmed checkpoint.
+
+Notably, the contract verifies consistency proofs of new merkle roots it
```
smeiklej, line 51: Same comment, should be Merkle.

smeiklej, line 51: Small thought: I think the higher-level readme we're planning is the right place to elaborate on the differences between using "normal" consistency proofs and using compact ranges, but it might be worth adding a quick note here saying that `RootWitness` and `CompactRangeWitness` are separate contracts that verify consistency in two different ways?  We can also wait to add this until we have the higher-level readme in place if that makes more sense.

smeiklej (approved):
LGTM modulo a few small comments about the readme.

witness/ethereum/README.md
sinasabet81: I updated this section; please take a look and let me know what you think!

sinasabet81:
@smeiklej just pushed an update with the fixes from your review; please take a look!

witness/ethereum/README.md
smeiklej: Looks good to me!

pavelkalinnikov (approved):

mhutchinson:
This PR has added a number of dependencies which are showing as having security vulns. Can someone take a look at https://github.com/google/trillian-examples/security/dependabot as a matter of urgency?

sinasabet81:
Sorry about that, looking into upgrading those versions in another PR now.

sinasabet81:
PR up bumping the package versions @ https://github.com/google/trillian-examples/pull/376

sinasabet81:
#367 seems to have fixed some of the issues, but other dependencies are still out of date; I think we're stuck here for now, since the eth-brownie repo seems to specify the minor version of these deps (eg. `urllib` in @ https://github.com/eth-brownie/brownie/blob/master/requirements.txt).

I'll follow up with the eth-brownie peeps and see if they can bump their dependencies, but even then we'll likely have to wait for them to cut a fresh release.

witness/ethereum/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+COPY ./tests /eth_wit/tests
+COPY Pipfile* /eth_wit/
+
+WORKDIR /eth_wit
+
+# Maybe a way to do this with determinism?
+# Does pip support verifying hashes of installed packages?
+RUN pip install pipenv
+RUN pipenv install --deploy --system
+RUN yarn global add ganache-cli
```
ernestognw, line 13: Adding global might have the same issues as with the `pip install`. I'm pretty sure you can lock to an specific ganache version by adding a `package.json`, the auto generated `yarn.lock` and running locally `yarn add ganache-cli`

Eventually if it's needed to run, it can be achieved with `npx ganache-cli` or `yarn run ganache-cli`

witness/ethereum/contracts/RootWitness.sol
```diff
@@ -0,0 +1,229 @@
...
+    ///   Corresponds to end index of the right range.
+    /// @param timestamp integer timestamp of the update.
+    /// @param signature raw signature over the marshalled bytes of new root.
+    /// @dev merge rightRange into compactRange.
+    /// @return treeData updated treeData entry for treeId.
+    function update(
+        bytes32 treeId,
+        bytes32[] calldata consistencyProof,
+        uint256 size,
+        uint256 timestamp,
```
ernestognw, line 98: Are the timestamps possibly optimizable in terms of size?

A `uint32` gives enough room for Unix epoch timestamps in case this is the format desired. Otherwise `uint40` is more than enough.

witness/ethereum/contracts/CompactRangeWitness.sol
```diff
@@ -0,0 +1,251 @@
...
+        bytes32 signature
+    );
+
+    address public owner;
+    /// @dev mapping from treeId to that tree's data.
+    mapping(bytes32 => TreeData) public treeDatas;
+
+    constructor() {
+        owner = msg.sender;
+    }
```
ernestognw, line 48: I guess this `Ownable` logic can be moved to another contract, it's repeated in both witness

witness/ethereum/contracts/CompactRangeWitness.sol
```diff
@@ -0,0 +1,251 @@
...
+    /// @dev merge rightRange into compactRange.
+    /// @return newRoot the new merkle root.
+    function update(
+        bytes32 treeId,
+        bytes32[] memory rightRange,
+        uint256 size,
+        uint256 timestamp,
+        bytes32 signature
+    ) external returns (bytes32 newRoot) {
+        require(owner == msg.sender, "msg.sender is not owner");
```
ernestognw, line 81: Maybe change to a `modifier`

ernestognw (commented):
Hi @sinasabet81

I just left some small nits to my understanding.

Great to see something like this

PR #364, opened 2021-06-29
Tweak leaf validator

AlCutter:
Complain if we find:
- a file in the log directory, but outside of the `leaves/pending` directory
- pending leaves AND modifications outside the log directory

### DIFF
--- a/serverless/deploy/github/leaf_validator/entrypoint.sh
+++ b/serverless/deploy/github/leaf_validator/entrypoint.sh
@@ -3,6 +3,10 @@
 # that this action would run against PRs which are effectively "queuing" leaves
 # and return success/failure depending on whether the leaves present in the PR
 # conform to a given set of requirements.
+#
+# We'll consider a PR good if either:
+# - it only touches files outside the log directory, or
+# - it only touches files in the log's leaves/pending directory.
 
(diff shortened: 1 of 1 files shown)


serverless/deploy/github/leaf_validator/entrypoint.sh
```diff
@@ -21,22 +25,35 @@ function main {
...
+                has_log_non_pending_files=1
+            else
+                echo "Found non-log file ${f}"
+                has_non_log_files=1
+            fi
     done <<< ${FILES}
 
-    if [[ ${is_bad} -ne 0 ]]; then
-        echo "::error::Found one or more invalid leaves in PR"
+    echo $has_non_log_files $has_log_pending_files $has_log_non_pending_files
```
mhutchinson, line 47: This is pretty raw output. Some kind of "::debug:" string like above seems more appropriate?

mhutchinson (approved):

serverless/deploy/github/leaf_validator/entrypoint.sh
AlCutter: Whoops, that shouldn't have been there - removed

PR #365, opened 2021-06-29
[serverless] Add a FetchLeafHashes func

AlCutter:


mhutchinson (approved):
Not for this PR now, but a) tests, b) renaming `nodeCache` are the obvious things that would make this change sweeter to review.

PR #373, opened 2021-06-30
[serverless] Add context param to long running/client methods.

AlCutter:


mhutchinson (approved):
The `New.*` methods taking in contexts is a bit surprising, but it transpires they are actually making requests within the factory method so this can't be avoided.

PR #376, opened 2021-06-30
ethwit - updating package versions

sinasabet81:
This subdir only has one explicit dep-- `eth-brownie`. I ran `pipenv update eth-brownie`, which should also update brownie's deps. Hopefully brownie's latest version has new enough versions of these deps to make dependabot stop complaining 😄


mhutchinson (approved):
I'm no expert on python's module versioning, but this is touching the deps that had vulnerabilities so let's try it and see. Thanks for doing this.

pavelkalinnikov (approved):

PR #378, opened 2021-07-02
[serverless] Add a couple of helpers to client

AlCutter:
- Expose func to fetch and open a checkpoint.
- Add utility func to check consistency between a number of checkpoints.


serverless/client/client.go
```diff
@@ -386,3 +389,38 @@ func (lst *LogStateTracker) Update(ctx context.Context) error {
...
+	lv := logverifier.New(h)
+
+	// Go through list of checkpoints pairwise, checking consistency
+	var a, b log.Checkpoint
+	for i := 0; i < len(cp)-1; i, a, b = i+1, cp[i], cp[i+1] {
+		if a.Size == b.Size {
+			if bytes.Equal(a.Hash, b.Hash) {
+				continue
+			}
+			return fmt.Errorf("two checkpoints with same size but different hashes")
```
mhutchinson, line 412: Including the hashes & size in this message would be cool.

serverless/client/client.go
```diff
@@ -386,3 +389,38 @@ func (lst *LogStateTracker) Update(ctx context.Context) error {
 	lst.LatestConsistentRaw, lst.LatestConsistent = cRaw, *c
 	return nil
 }
+
+// CheckConsistency is a wapper function which simplifies verifying consistency between two or more checkpoints.
+func CheckConsistency(ctx context.Context, h hashers.LogHasher, f Fetcher, cp []log.Checkpoint) error {
```
mhutchinson, line 394: I wonder if having `cp` being variadic would be useful in some cases?

serverless/client/client.go
```diff
@@ -386,3 +389,38 @@ func (lst *LogStateTracker) Update(ctx context.Context) error {
...
+			return fmt.Errorf("two checkpoints with same size but different hashes")
+		}
+		if a.Size > 0 {
+			cp, err := pb.ConsistencyProof(ctx, a.Size, b.Size)
+			if err != nil {
+				return fmt.Errorf("failed to fetch consistency between sizes %d, %d: %v", a.Size, b.Size, err)
+			}
+			if err := lv.VerifyConsistencyProof(int64(a.Size), int64(b.Size), a.Hash, b.Hash, cp); err != nil {
+				return fmt.Errorf("invalid consistency proof between sizes %d, %d: %v", a.Size, b.Size, err)
+
```
mhutchinson, line 421: nit blank line

serverless/client/client.go
```diff
@@ -47,7 +48,9 @@ import (
 // based implementation MUST return this error when it receives a 404 StatusCode.
 type Fetcher func(ctx context.Context, path string) ([]byte, error)
 
-func fetchCheckpointAndParse(ctx context.Context, f Fetcher, v note.Verifier) (*log.Checkpoint, []byte, error) {
+// FetchCheckpoint retrieves and opens a checkpoint from the log.
+// Returns both the parsed structure and the raw serialised checkpoint.
+func FetchCheckpoint(ctx context.Context, f Fetcher, v note.Verifier) (*log.Checkpoint, []byte, error) {
```
mhutchinson, line 53: Have you tried this on a checkpoint that has been co-signed? `note.Open` will return an error if there exist any signatures that don't check out. It may be worth noting this in the description. It could even be a TODO. Eventually this should take the log sig verifier, and parse providing that matches. It could also return (either as another argument, or in an error) the list of unverified sigs.

mhutchinson (approved):

serverless/client/client.go
AlCutter: done

serverless/client/client.go
AlCutter: Hmm, dunno, do you have any cases like that in mind?

serverless/client/client.go
AlCutter: ding'd

serverless/client/client.go
AlCutter: I don't think that's true...  `note.Open` will err if _none_ of the sigs are valid or we didn't pass in a matching `Verifier` (https://play.golang.org/p/U6FJqvl0ZkD)

My thinking here is that this is explicitly fetching a checkpoint from a specific log - so it MUST be signed by the log, anything else (e.g. witness sigs) is a concern for somewhere else.  There might be a better construction which we can move to, but I was trying to strike a balance between convenience and stopping folks from getting it too wrong (e.g. doing a note.Open with {logA, Witness} verifiers on a checkpoint signed by {LogE, Witness} would open without error but is probably not what you would expect here)

PR #379, opened 2021-07-02
[serverless] Need at least 2 checkpoints to verify consistency

AlCutter:


mhutchinson (commented):
I assumed this was a feature in the previous PR. I'd say a single checkpoint is consistent with itself. Is there a specific reason this change is required?

AlCutter:
It'll currently panic if you pass in a single CP via the slice, I figured this is a nicer failure mode...

mhutchinson (approved):

PR #396, opened 2021-07-09
[serverless] Add an action to handle witness PRs

AlCutter:


serverless/deploy/github/witness_pr/README.md
```diff
@@ -0,0 +1,84 @@
+# witness_pr
+
+`witness_pr` is a GitHub Action for handling incoming PRs from _witnesses_ containing a
+co-signed log checkpoint which the witness has verified is consistent with earlier
+log checkpoints it has seen.
+
+This action would be used by a serverless log which was participating in an ecosystem
```
mhutchinson, line 7: This presents the case as a binary thing:
 0 - witnesses not addressable; logs bundle signatures
 1 - witnesses addressable, logs don't bundle signatures

I think the log & witness issues are somewhat orthogonal, though the situation that both don't have any features is equivalent to no witnessing. TL;DR: there's a third case which may also exist: (perhaps some, maybe all) witnesses are addressable, and the log also bundles sigs as a convenience to clients.

serverless/deploy/github/witness_pr/README.md
```diff
@@ -0,0 +1,84 @@
...
+The `witness_pr` action requires a copy of the PR branch, as well as a pristine checkout of
+the log's `master` (or `main`) branch.
+Public keys for known witnesses should already be committed on the `master` branch.
+
+The `witness_pr` action:
+1. checks that there is only one modified file present in the PR: `checkpoint.witnessed`
+2. attempts to combine signatures from the PR's `checkpoint.witnessed` file with an existing
+   `checkpoint.witnessed` file in the `master` branch, and updates the file in the PR with
+   the union of the signatures.
+3. if step (2) failed (e.g. because the log published an updated `checkpoint` file which the
```
mhutchinson, line 21: The explanation of what could cause this path is only clear to me because I was involved in discussion of this feature workflow. One thing to consider is an overview section above which lays out the `c` and `c.w` files, with the invariant that `c >= c.w` (i.e. the witness file can never be newer). The expected workflow for witnesses is to simply follow the `c` file, verify the evolution, then submit their signature as `c.w`.

Given the above, I think that sets the stage for realizing that there is a race condition between new checkpoints being generated by the log, and the witnesses signing them and submitting them back.

serverless/deploy/github/witness_pr/README.md
```diff
@@ -0,0 +1,84 @@
...
+    # Checkout the log's master branch into pristine/
+    - uses: actions/checkout@v2
+      with:
+        path: 'pristine'
+        ref: 'master'
+        fetch-depth: 0
+    # Attempt to combine the signatures on the checkpoint in the PR with the log's latest checkpoint/checkpoint.witnessed file
+    - name: Validate and combine
+      id: validate_combine
+      uses: AlCutter/trillian-examples/serverless/deploy/github/witness_pr@serverless_recipe_witness_pr
```
mhutchinson, line 69:
```suggestion
      uses: google/trillian-examples/serverless/deploy/github/witness_pr@serverless_recipe_witness_pr
```

Obviously chicken and egg with this PR being merged, but in essence it's sound?

mhutchinson (approved):

serverless/deploy/github/witness_pr/README.md
AlCutter: Yeah, this was really just to give a bit of context for how this action is currently intended to be used - I expect there'll be a bunch of refinement as we experiment with different scenarios (and maybe we can doc those in a README under `/witness` or something?)

serverless/deploy/github/witness_pr/README.md
AlCutter: I've added an overview chunk - PTAL?

serverless/deploy/github/witness_pr/README.md
AlCutter: ta, fixed up to where it'll land RSN :)

serverless/deploy/github/witness_pr/README.md
```diff
@@ -0,0 +1,109 @@
...
+`witness_pr` is a GitHub Action for handling incoming PRs from _witnesses_ containing a
+co-signed log checkpoint which the witness has verified is consistent with earlier
+log checkpoints it has seen.
+
+This action would be used by a serverless log which was participating in an ecosystem
+with non-addressable witnesses.
+
+## Overview
+
+The serverless log creates a `checkpoint` file as part of the integration step, this file
```
mhutchinson, line 12:
```suggestion
The serverless log creates a `checkpoint` file as part of the integration step. This file
```

serverless/deploy/github/witness_pr/README.md
mhutchinson: LG, thanks.

mhutchinson (approved):

PR #397, opened 2021-07-12
[witness] adding support for compact ranges

smeiklej:


witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -219,6 +244,52 @@ func (w *Witness) getLatestChkpt(queryRow func(query string, args ...interface{}
...
+}
+
+// verifyRange verifies the new checkpoint against the stored and given compact
+// range and outputs the new compact range if verification succeeds.
+func (w *Witness) verifyRange(chkpt *log.Checkpoint, oldSize uint64, h hashers.LogHasher, curRangeRaw [][]byte, newRangeRaw [][]byte) ([][]byte, error) {
+	rf := compact.RangeFactory{Hash: h.HashChildren}
+	curRange, err := rf.NewRange(0, oldSize, curRangeRaw)
+	if err != nil {
+		return nil, fmt.Errorf("can't form current compact range: %v", err)
+	}
```
AlCutter, line 276: Worth doing a `GetRootHash` on curRange and double checking it's valid for the old chkpt? (would help debugging in the case that updates were failing because the range in the DB was corrupt rather than the incoming proof being wrong, etc.)

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -120,9 +126,8 @@ func (w *Witness) GetLogs() ([]string, error) {
 	return logs, nil
 }
 
-// GetCheckpoint gets a checkpoint for a given log, which will be
-// consistent with all other checkpoints for the same log.  It also signs it
-// under the witness' key.
+// GetCheckpoint gets a checkpoint for a given log, which is consistent with all
+// other checkpoints for the same log.  It also signs it under the witness' key.
```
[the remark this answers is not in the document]
smeiklej: Sure, done.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -167,11 +172,21 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
 	if err != nil {
 		return 0, fmt.Errorf("couldn't create tx: %v", err)
 	}
-	latest, err := w.getLatestChkpt(tx.QueryRow, logID)
+	// Get either just the checkpoint or the checkpoint and compact range.
+	var latest Chkpt
+	var curRange [][]byte
+	if logInfo.UseCompact {
+		latest, curRange, err = w.getLatestChkptAndRange(tx.QueryRow, logID)
```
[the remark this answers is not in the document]
smeiklej: Yeah, I knew this wasn't so nice :).  I've done this now, which turned into a slightly bigger change because I also took the opportunity to strip `size` from the database (it was already unnecessary since we haven't been storing multiple checkpoints for a while) and get rid of the `Chkpt` struct as a result too. 

Anyway both either take in or return raw `[]byte` arguments now and it's up to the caller to (un)marshal.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -185,18 +200,28 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
 		// Complain if latest is bigger than chkpt.
 		return p.Size, fmt.Errorf("cannot prove consistency backwards (%d < %d)", chkpt.Size, p.Size)
 	case chkpt.Size > p.Size:
-		// Potentially a valid update
-		if err := logInfo.LogV.VerifyConsistencyProof(int64(p.Size), int64(chkpt.Size), p.Hash, chkpt.Hash, proof); err != nil {
-			// Complain if the checkpoints aren't consistent.
-			return p.Size, fmt.Errorf("failed to verify consistency proof: %v", err)
+		// Potentially a valid update, use either plain consistency proofs
+		// or compact ranges to verify, depending on the log.
+		if !logInfo.UseCompact {
```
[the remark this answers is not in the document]
smeiklej: Done, switched to `switch` per the comment below anyway.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -185,18 +200,28 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+			logV := logverifier.New(logInfo.Hasher)
+			if err := logV.VerifyConsistencyProof(int64(p.Size), int64(chkpt.Size), p.Hash, chkpt.Hash, proof); err != nil {
+				// Complain if the checkpoints aren't consistent.
+				return p.Size, fmt.Errorf("failed to verify consistency proof: %v", err)
+			}
+			// If the consistency proof is good we store chkptRaw.
+			return p.Size, w.setCheckpoint(tx, logID, c)
+		}
+		newRange, err := w.verifyRange(chkpt, p.Size, logInfo.Hasher, curRange, proof)
```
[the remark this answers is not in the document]
smeiklej: Changed to `switch`!

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -185,18 +200,28 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+		if err != nil {
+			return p.Size, fmt.Errorf("failed to verify consistency: %v", err)
```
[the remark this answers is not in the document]
smeiklej: Fixed :).

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: Yup, good idea!  It's also currently possible that for the first checkpoint the caller gives complete garbage for the range, which we wouldn't find out until we went to use it.  So I guess it would be good to also add a validity check before storing it (this is line 178 of `witness.go`, where currently I just store it without any checks).

AlCutter: 👍

smeiklej: Okay, that's done now!

witness/golang/cmd/witness/example.conf
```diff
@@ -2,6 +2,7 @@
    "logs" : [
       {
          "hashstrategy" : "default",
+	 "usecompact" : "false",
```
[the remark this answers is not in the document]
smeiklej: Cool, fixed the three issues.  I'll probably wait to change to a more general `proofType` until it seems like that's needed?

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: This is done now!

witness/golang/cmd/witness/example.conf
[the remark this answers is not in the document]
AlCutter: Yeah, that's probably the right call.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -258,27 +267,32 @@ func (w *Witness) verifyRange(chkptNew *log.Checkpoint, chkptOld *log.Checkpoint
...
+func (w *Witness) verifyRangeHash(rootHash []byte, rng *compact.Range) error {
+	h, err := rng.GetRootHash(nil)
 	if err != nil {
-		return nil, fmt.Errorf("can't get root hash for range: %v", err)
+		return fmt.Errorf("can't get root hash for range: %v", err)
 	}
-	if !bytes.Equal(hashNew, chkptNew.Hash) {
-		return nil, fmt.Errorf("hashes aren't equal (got %x, given %x)", hashNew, chkptNew.Hash)
+	if !bytes.Equal(rootHash, h) {
+		return fmt.Errorf("old hashes aren't equal (got %x, given %x)", h, rootHash)
```
AlCutter, line 293: Probably don't want `old` here?

AlCutter (approved):
just one nit inline

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
 	// with respect to older checkpoints.  Bind this all in a transaction to
 	// avoid race conditions when updating the database.
 	tx, err := w.db.BeginTx(ctx, nil)
 	if err != nil {
 		return 0, fmt.Errorf("couldn't create tx: %v", err)
 	}
-	latest, err := w.getLatestChkpt(tx.QueryRow, logID)
+	// Get the latest checkpoint and compact range (if one exists).
+	latestRaw, rangeRaw, err := w.getLatestChkptData(tx.QueryRow, logID)
 	if err != nil {
```
pavelkalinnikov, line 170: The size and the complexity of the `Update` function is significant, consider splitting it. For example, I feel like the compact range and non-(compact range) flows can be split into different functions.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+		return 0, fmt.Errorf("couldn't parse stored checkpoint: %v", err)
+	}
+	// Parse the compact range if we're using one.
+	var curRange log.Proof
+	if logInfo.UseCompact {
+		if err := curRange.Unmarshal(rangeRaw); err != nil {
+			return 0, fmt.Errorf("couldn't unmarshal proof: %v", err)
+		}
 	}
 	switch {
```
pavelkalinnikov, line 205: There is one dominating `case` in this `switch`. I think it would be more readable if the error "new < old" case and the fast-path "new == old" case were put in the beginning as two `if`s, and then the dominating "new > old" case was left as a remaining unconditional flow (with the benefit of not needing to indent a big chunk of code). You wouldn't need the `default` boilerplate too.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
-		if err := logInfo.LogV.VerifyConsistencyProof(int64(p.Size), int64(chkpt.Size), p.Hash, chkpt.Hash, proof); err != nil {
-			// Complain if the checkpoints aren't consistent.
-			return p.Size, fmt.Errorf("failed to verify consistency proof: %v", err)
+	case chkptNew.Size < chkptOld.Size:
+		// Complain if old checkpoint is bigger than new one.
+		return chkptOld.Size, fmt.Errorf("cannot prove consistency backwards (%d < %d)", chkptNew.Size, chkptOld.Size)
+	case chkptNew.Size > chkptOld.Size:
+		// Potentially a valid update, use either plain consistency proofs
+		// or compact ranges to verify, depending on the log.
+		switch {
```
pavelkalinnikov, line 212: You don't need a `switch` here, just `if-else`? Also, because the first branch ends with a `return`, you don't need `else`, if can be just put unconditionally after the `if` branch.

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: Nope, good catch!

pavelkalinnikov (commented):
A few readability concerns. Reading further.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
-		// Complain if latest is bigger than chkpt.
-		return p.Size, fmt.Errorf("cannot prove consistency backwards (%d < %d)", chkpt.Size, p.Size)
-	case chkpt.Size > p.Size:
-		// Potentially a valid update
-		if err := logInfo.LogV.VerifyConsistencyProof(int64(p.Size), int64(chkpt.Size), p.Hash, chkpt.Hash, proof); err != nil {
-			// Complain if the checkpoints aren't consistent.
-			return p.Size, fmt.Errorf("failed to verify consistency proof: %v", err)
+	case chkptNew.Size < chkptOld.Size:
+		// Complain if old checkpoint is bigger than new one.
+		return chkptOld.Size, fmt.Errorf("cannot prove consistency backwards (%d < %d)", chkptNew.Size, chkptOld.Size)
```
pavelkalinnikov, line 208: It's conventional in Go to return zero values in all error cases. Applies here and throughout this file.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
 		}
-		return Chkpt{}, err
+		return nil, nil, err
+	}
+	return rawChkpt, rawPf, nil
+}
+
+// verifyRange verifies the new checkpoint against the stored and given compact
+// range and outputs the new compact range if verification succeeds.
+func (w *Witness) verifyRange(chkptNew *log.Checkpoint, chkptOld *log.Checkpoint, h hashers.LogHasher, curRangeRaw [][]byte, newRangeRaw [][]byte) ([][]byte, error) {
```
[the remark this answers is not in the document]
pavelkalinnikov: nit: Be consistent with either "old" or "cur"? Or consider `prev`/`next` too ("new" is a reserved word in Go, so it can be annoying if you want to create variables `old` and `new`, for instance).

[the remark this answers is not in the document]
pavelkalinnikov: From the code below it seems that the "new range" is more like a "delta"/"diff" range between the two checkpoints?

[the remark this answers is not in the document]
pavelkalinnikov: nit: Consider removing "Raw" from the names. I know you want to distinguish `curRange` from `curRangeRaw`, but the variable 2 lines below can be named even shorter like `r` because it's a local short-living variable (and short names are ok in Go).

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+	rf := compact.RangeFactory{Hash: h.HashChildren}
+	curRange, err := rf.NewRange(0, chkptOld.Size, curRangeRaw)
+	if err != nil {
+		return nil, fmt.Errorf("can't form current compact range: %v", err)
+	}
+	// As a sanity check, make sure the old checkpoint is consistent with
+	// the current range.
```
pavelkalinnikov, line 269: What is the distinction between "old" and "current" here? Should we reduce this to only one (or zero?) adjective? E.g.

`// Make sure that compact range is consistent with the checkpoint.`

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+	curRange.AppendRange(newRange, nil)
+	if err := w.verifyRangeHash(chkptNew.Hash, curRange); err != nil {
+		return nil, fmt.Errorf("new root hash doesn't verify: %v", err)
+	}
+	return curRange.Hashes(), nil
+}
+
+// verifyRangeHash computes the root hash of the compact range and compares it
+// against the one given as input, returning an error if they aren't equal.
+func (w *Witness) verifyRangeHash(rootHash []byte, rng *compact.Range) error {
```
pavelkalinnikov, line 287: This doesn't need to be a method of `w`.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+			if logInfo.UseCompact {
+				rf := compact.RangeFactory{Hash: logInfo.Hasher.HashChildren}
+				rng, err := rf.NewRange(0, chkptNew.Size, proof)
+				if err != nil {
+					return 0, fmt.Errorf("can't form compact range: %v", err)
+				}
+				if err := w.verifyRangeHash(chkptNew.Hash, rng); err != nil {
+					return 0, fmt.Errorf("input root hash doesn't verify: %v", err)
+				}
+				rangeRaw := []byte(log.Proof(proof).Marshal())
```
pavelkalinnikov, line 186: Does the range need to be stored in the marshalled format? Can it be just an array of byte arrays?

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+	curRange, err := rf.NewRange(0, chkptOld.Size, curRangeRaw)
+	if err != nil {
+		return nil, fmt.Errorf("can't form current compact range: %v", err)
+	}
+	// As a sanity check, make sure the old checkpoint is consistent with
+	// the current range.
+	if err := w.verifyRangeHash(chkptOld.Hash, curRange); err != nil {
+		return nil, fmt.Errorf("old root hash doesn't verify: %v", err)
+	}
+	newRange, err := rf.NewRange(chkptOld.Size, chkptNew.Size, newRangeRaw)
```
pavelkalinnikov, line 273: naming nit: `delta`?

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,75 +154,153 @@ func (w *Witness) Update(ctx context.Context, logID string, chkptRaw []byte, pro
...
+	if err != nil {
+		return nil, fmt.Errorf("can't form new compact range: %v", err)
```
pavelkalinnikov, line 275: delta/diff?

pavelkalinnikov (commented):
Some more comments.

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: Changed!

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: Sure, changed to `delta`.

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: I went with one adjective I think ("stored").

witness/golang/cmd/witness/internal/witness/witness.go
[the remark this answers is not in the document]
smeiklej: Okay, I did the following:

- Renamed chkptOld and chkptNew to prev and next, as suggested.  Called their "raw" counterparts prevRaw and nextRaw, because I do think this is helpful (for me if no one else).
- Called the corresponding ranges prevRange and nextRange, where they were long-lived and it made sense to do so.  Otherwise used variable names like r and rng.
- Made `verifyRange` just be a standalone function.
- Renamed the "new" range to deltaRange.
- Removed almost all uses of old, cur, and new, so it should be consistent now.

witness/golang/cmd/witness/internal/witness/witness.go
pavelkalinnikov: YAGNI? It would make sense if this PR contained the use-case for this, but it doesn't. Also, conventionally, if anything returns `err != nil`, the handling code doesn't try interpreting the other return values and forwards the error up the stack / prints it / etc.

Another thing is, `Update` can return an error before the previous size was read. How does the caller know whether the returned 0 is indeed the previous size, or it's just that `Update` has failed to read it? In order to distinguish these 2 cases the client would need to parse the error further.

PR #398, opened 2021-07-14
[Serverless] Add one-shot witness PR creation tool

AlCutter:
This PR adds a companion tool to the witness PR handler action in #396 .

### DIFF
--- a/serverless/cmd/feeder/feed-to-github
+++ b/serverless/cmd/feeder/feed-to-github
@@ -25,10 +25,26 @@ EOF
   exit 1
 }
 
+function checkExist() {
+  local missing=()
+  for i in $*; do
+    if ! type ${i} &> /dev/null; then
+      missing+=("${i}")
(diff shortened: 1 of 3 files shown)


serverless/cmd/feeder/feed-to-github
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+#
+# feed-to-github is a wrapper around the serverless feeder command which does a
```
mhutchinson, line 3: Should this have a `.sh` extension?

mhutchinson (approved):

serverless/cmd/feeder/feed-to-github
AlCutter: Funnily enough it did have one originally, and then I read [this](https://google.github.io/styleguide/shellguide.html#file-extensions) so I removed it :)

PR #399, opened 2021-07-14
Support writing & reading checkpoints in DB

mhutchinson:
Checkpoints are stored with the compact range, which allows the log contents to be verified incrementally.

Verified checkpoints are now stored by the `ctverify` tool. Upon verifying a new checkpoint, the latest verified checkpoint will be loaded and only the leaves added since that checkpoint will directly be checked.

This is a performance improvement for verification, and it also provides tooling that reads from the log (e.g. map building code) to find the log checkpoints. This code can either trust the local DB hasn't been modified since verification, or it can re-verify all leaves against the checkpoint if desired. In the case of the map, the log checkpoint used will be embedded within the map checkpoint in order to complete an audit trail.

### DIFF
--- a/clone/logdb/database.go
+++ b/clone/logdb/database.go
@@ -16,16 +16,12 @@
 package logdb
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
+	"encoding/gob"
 	"errors"
 	"fmt"
-)
-
-const (
-	// This can be changed to a field of Database or metadata in the database if
-	// we need to support other sizes in the future.
-	hashSizeBytes = 32
 )
 
 // ErrNoDataFound is returned when the DB appears valid but has no data in it.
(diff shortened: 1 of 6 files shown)


clone/logdb/database.go
```diff
@@ -40,18 +46,78 @@ func NewDatabase(connString string) (*Database, error) {
...
+		}
+		return 0, nil, nil, fmt.Errorf("Scan(): %v", err)
+	}
+	compactRange = make([][]byte, len(srs)/hashSizeBytes)
+	for i := range compactRange {
+		from := i * hashSizeBytes
+		to := from + hashSizeBytes
+		compactRange[i] = srs[from:to]
+	}
+	return
```
AlCutter, line 118: Perhaps you could use GOB encoding of the array rather than doing it manually?

AlCutter (approved):

clone/logdb/database.go
mhutchinson: Now done. Ta.

PR #405, opened 2021-07-15
Serverless fix actions

AlCutter:
- Add missing deps
- Add test to presubmit script to build Docker images.

### DIFF
--- a/scripts/presubmit.sh
+++ b/scripts/presubmit.sh
@@ -109,7 +109,7 @@ main() {
         ./...
     fi
 
-    echo "builing serverless actions"
+    echo "building serverless actions"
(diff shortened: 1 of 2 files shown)


mhutchinson (approved):

scripts/presubmit.sh
```diff
@@ -108,6 +108,11 @@ main() {
         -timeout=${GO_TEST_TIMEOUT:-5m} \
         ./...
     fi
+
+    echo "builing serverless actions"
```
mhutchinson, line 112:
```suggestion
    echo "building serverless actions"
```

AlCutter: gah, ta

PR #409, opened 2021-07-19
serverless: Fix consistency checks

pavelkalinnikov:
The consistency checking function ignores the first pair of checkpoints.
This change adds a test that failed, and fixes it.


AlCutter (approved):
Thanks!

PR #410, opened 2021-07-19
serverless: Add rehashing to consistency proofs

pavelkalinnikov:
This change adds a test for a bigger tree to showcase the failure,
and fixes the bug by introducing the `Rehash` step.


AlCutter (approved):
Thanks again Pavel!

PR #412, opened 2021-07-19
[witness] http server tests + small change to Update API

smeiklej:
Change Update to return the (signed) latest witness checkpoint.

This is to make it easier for entities who want to get their checkpoints witnessed and then distribute these witnessed checkpoints, as they no longer have to call Update and then GetCheckpoint.

Also:
- Makes Update PUT instead of POST.
- Added tests for the HTTP server.
- Refactoring in the witness and its test code.


witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -121,120 +121,113 @@ func (w *Witness) GetLogs() ([]string, error) {
...
 	}
 	if next.Size < prev.Size {
 		// Complain if prev is bigger than next.
-		return prev.Size, fmt.Errorf("cannot prove consistency backwards (%d < %d)", next.Size, prev.Size)
+		return prevRaw, status.Errorf(codes.Aborted, "cannot prove consistency backwards (%d < %d)", next.Size, prev.Size)
 	}
 	if next.Size == prev.Size {
 		if !bytes.Equal(next.Hash, prev.Hash) {
-			return prev.Size, fmt.Errorf("checkpoint for same size log with differing hash (got %x, have %x)", next.Hash, prev.Hash)
+			return prevRaw, status.Errorf(codes.Aborted, "checkpoint for same size log with differing hash (got %x, have %x)", next.Hash, prev.Hash)
```
AlCutter, line 191: nit: is this more `FAILED_PRECONDITION`(neither this CP's nor any later CP on the same timeline is ever going to be consistent with the witness' one)

Also, it's maybe worth logging the full CP as evidence?

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/google/trillian-examples/witness/golang/api"
+	"github.com/google/trillian-examples/witness/golang/cmd/witness/internal/witness"
+	"github.com/google/trillian/merkle/rfc6962/hasher"
+	"github.com/gorilla/mux"
+	_ "github.com/mattn/go-sqlite3" // Load drivers for sqlite3
```
AlCutter, line 33: Usually separate the anon imports into their own section below

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+	crNext  = []byte("Log Checkpoint v0\n15\nsrKoB8sjvP1QAt1Ih3nqGHzjvmtRLs/geQdehrUHvqs=\n\n— monkeys h74qVU+3T1AVNo+tAsISWzMwEDag8xOwQLSzxJELzHH9N/4cz6M+VbJ89Ku//gybRPPUFP/8Fcvxpqb6++ZyLk83oAE=\n")
+	crProof = [][]byte{
+		dh("ef626e0b64023948e57f34674c2574b3078c5af59a2faa095f4948736e8ca52e", 32),
+		dh("8f75f7d88d3679ac6dd5a68a81215bfbeafe8c566b93013bbc82e64295628c8b", 32),
+		dh("e034fb7af8223063c1c299ed91c11a0bc4cec15afd75e2abe4bb54c14d921ef0", 32),
+	}
+)
+
+type LogOpts struct {
```
AlCutter, line 63: Does this need to be exported?

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+	ID         string
+	PK         string
+	useCompact bool
+}
+
+func newOpts(d *sql.DB, logs []LogOpts) (witness.Opts, error) {
```
AlCutter, line 69: Worth taking `t *testing.T` here as in `newWitness`?

AlCutter (commented):
Few quick bits (will look more after lunch).

Might have been good to separate this into 2 smaller PRs if that was possible?

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+			chkpts:     [][]byte{mInit},
+			wantStatus: http.StatusOK,
+			wantBody:   `["monkeys"]`,
+		}, {
+			desc:       "two logs",
+			logIDs:     []string{"bananas", "monkeys"},
+			logPKs:     []string{bPK, mPK},
+			chkpts:     [][]byte{bInit, mInit},
+			wantStatus: http.StatusOK,
+			wantBody:   `["bananas","monkeys"]`,
```
AlCutter, line 157: I wonder if it might be less brittle to have this be `wantResponse struct ...` which gets compared to a similar struct that the JSON response is parsed into?
As it is, any changes to whitespace formatting would cause the test to fail, even though it's a valid/correct response.

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+			if got, want := resp.StatusCode, test.wantStatus; got != want {
+				t.Errorf("status code got != want (%d, %d)", got, want)
+			}
+			if len(test.wantBody) > 0 {
+				body, err := ioutil.ReadAll(resp.Body)
+				if err != nil {
+					t.Errorf("failed to read body: %v", err)
+				}
+				if got, want := string(body), test.wantBody; got != test.wantBody {
+					t.Errorf("got '%s' want '%s'", got, want)
```
AlCutter, line 200: /'%s'/%q/

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+			defer ts.Close()
+
+			client := ts.Client()
+			url := fmt.Sprintf("%s%s", ts.URL, api.HTTPGetLogs)
+			resp, err := client.Get(url)
+			if err != nil {
+				t.Errorf("error response: %v", err)
+			}
+			if got, want := resp.StatusCode, test.wantStatus; got != want {
+				t.Errorf("status code got != want (%d, %d)", got, want)
```
AlCutter, line 192: nit: might be nicer to read `got %d, want %d`

witness/golang/cmd/witness/internal/http/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+				if _, err := w.Update(ctx, test.setID, test.c, nil); err != nil {
+					t.Errorf("failed to set checkpoint: %v", err)
+				}
+			}
+			// Now set up the http server.
+			r := mux.NewRouter()
+			server := NewServer(w)
+			server.RegisterHandlers(r)
+			ts := httptest.NewServer(r)
+			defer ts.Close()
```
AlCutter, line 260: Is this identical to the setup in the previous test? is it worth extracting to a `createTestEnv(...) (*httptest.Server, func())` method or similar?

witness/golang/cmd/witness/example.conf
```diff
@@ -2,7 +2,7 @@
     "logs": [
         {
             "hashStrategy": "default",
-            "logID": "testlog",
+            "logID": "monkeys",
```
mhutchinson, line 5: nit: this is fun, but may be a head scratcher for serious people looking at this example config :-) `examplelog` or `mylogid` would be the safe paths.

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -121,120 +121,113 @@ func (w *Witness) GetLogs() ([]string, error) {
...
-	return cosigned, nil
+	return chkpt, nil
 }
 
 // Update updates the latest checkpoint if nextRaw is consistent with the current
-// latest one for this log. It returns the latest checkpoint size before the
-// update was applied (or just what was fetched if the update was unsuccessful).
-func (w *Witness) Update(ctx context.Context, logID string, nextRaw []byte, proof [][]byte) (uint64, error) {
+// latest one for this log. If the update was successful it returns a cosigned
+// version of nextRaw and if not it returns its (signed) latest checkpoint.
```
mhutchinson, line 135:
```suggestion
// latest one for this log. This returns the witness's latest cosigned checkpoint,
// which will be a signed version of nextRaw if this update was applied.
```

I think the important property is that whatever it returns is whatever the next call to update should use as its base (unless the world changes again). I think my suggestion is a little clearer on that, but I'm not attached to it.

witness/golang/cmd/witness/internal/witness/witness.go
mhutchinson: +1. The difference is subtle at some level, but this isn't the kind of thing clients should just think they need to casually "retry at a higher level".

witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -301,3 +294,25 @@ func (w *Witness) setChkptData(tx *sql.Tx, logID string, c []byte, rng []byte) e
 		logID, c, rng)
 	return tx.Commit()
 }
+
+// setInitChkptData stores the data for an initial checkpoint and, if using one,
+// its associated compact range.
+func (w *Witness) setInitChkptData(tx *sql.Tx, logID string, c *log.Checkpoint, cRaw []byte, rngRaw [][]byte) error {
```
mhutchinson, line 300: The refactoring of this method would be better in a different PR. It took me a while to realize we weren't making changes to the handling of compact ranges when I saw this deleted in the other place.

witness/golang/cmd/witness/internal/http/server_test.go
smeiklej: Okie, done.

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: Okay, that's a nice way of thinking about it.  Used your text mostly verbatim.

witness/golang/cmd/witness/internal/witness/witness.go
smeiklej: Yup, agreed - sorry for the tangled mess!  This is what happens apparently when I am going on holidays and try to rush things :/.

witness/golang/cmd/witness/internal/http/server_test.go
smeiklej: Sure, changed to unmarshalling JSON and comparing the two lists.

smeiklej (commented):
Thanks for the comments!  Think they're all addressed now.

mhutchinson (commented):
Thanks for the changes. My main suggestion is to put a few minutes work into the commit / PR description. I think something like:

```
Change Update to return the latest witness checkpoint.

1-line justification for this change (saves roundtrips).

Also:
 * Makes Update PUT instead of POST
 * Added HTTP tests
 * Refactoring
```

mhutchinson (approved):
This looks good to me, but honestly, I'm not very good at reviewing unfocused PRs. Please let Al approve too :-)

witness/golang/cmd/witness/internal/witness/witness.go
AlCutter: Ah - I meant literally logging the CP (i.e. `glog.Warningf("Oh noes: %s", cpRaw)`) since it's potentially signed evidence of a log misbehaving.

AlCutter (approved):
LGTM (feel free to do the logging thing either in this PR or as a follow-up if you think it's worth adding in)

PR #413, opened 2021-07-20
Update trillian to 785d9ec

AlCutter:


binary_transparency/firmware/cmd/ft_personality/impl/ft_personality.go
```diff
@@ -33,8 +33,8 @@ import (
 	"github.com/gorilla/mux"
 	"golang.org/x/mod/sumdb/note"
 
-	_ "github.com/google/trillian/merkle/rfc6962/hasher" // Load hashers
-	_ "github.com/mattn/go-sqlite3"                      // Load drivers for sqlite3
+	_ "github.com/google/trillian/merkle/rfc6962" // Load hashers
```
pavelkalinnikov, line 36: This is not needed as we don't have "registers" now?

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/gorilla/mux"
 	"golang.org/x/mod/sumdb/note"
 
-	_ "github.com/google/trillian/merkle/rfc6962/hasher" // Load hashers
+	_ "github.com/google/trillian/merkle/rfc6962" // Load hashers
```
pavelkalinnikov, line 32: Same here.

serverless/client/client.go
```diff
@@ -116,7 +116,7 @@ func NewProofBuilder(ctx context.Context, cp log.Checkpoint, h compact.HashFn, f
 // This function uses the passed-in function to retrieve tiles containing any log tree
 // nodes necessary to build the proof.
 func (pb *ProofBuilder) InclusionProof(ctx context.Context, index uint64) ([][]byte, error) {
-	nodes, err := merkle.CalcInclusionProofNodeAddresses(int64(pb.cp.Size), int64(index), int64(pb.cp.Size))
+	nodes, err := merkle.CalcInclusionProofNodeAddresses(int64(pb.cp.Size), int64(index))
```
pavelkalinnikov, line 119: Needs `merkle.Rehash`, similarly to how it's in `ConsistencyProof`.

binary_transparency/firmware/cmd/ft_personality/impl/ft_personality.go
AlCutter: Good spot - these were all done mechanically with `gofmt -r`

binary_transparency/firmware/cmd/ft_witness/impl/ft_witness.go
AlCutter: done

serverless/client/client.go
AlCutter: aha, ta - this was unnecessary previously because of the logic which was previously inside `CalcInclusionProofNodeAddresses`?

pavelkalinnikov: Right. Previously, it would return a different set of nodes if the 2 tree sizes are the same. Now it always returns non-ephemeral nodes and needs rehashing. With a future https://github.com/google/trillian/pull/2576 it will be possible to still workaround and get the ephemeral node, with a bit more conscious effort.

AlCutter: 👍

pavelkalinnikov (approved):
LGTM as long as the CI is happy.

PR #414, opened 2021-07-20
[Serverless] Example of log compiled to WASM and running in browser

AlCutter:
Uses the browser's `sessionStorage` for log state.

Intended mainly as an experiment to see if we can make interactive in-browser demos easily, but might be useful as the underpinnings for a "Log Explorer" for serverless logs hosted on GitHub etc.?

See instructions in the `serverless/wasm/main.go` file on how to spin up and try at home.


serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build wasm
+
+// Package webstorage provides a simple log storage implementation based on webstorage.
+// It only really makes sense for wasm targets.
```
mhutchinson, line 18: I think the word Javascript would be good to add into here somewhere. The Go ecosystem seems to be favouring doc view (over code view) more, and I can easily see some developers (i.e. me if I hadn't had certain non-work conversations in my life) not knowing quite what this implies. It would be easy to assume `webstorage` was talking about putting stuff in google cloud, or github, or something.

serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+	v := getStorage().Call("getItem", js.ValueOf(k))
+	if js.Null().Equal(v) {
+		return nil, os.ErrNotExist
+	}
+	return []byte(v.String()), nil
+}
+
+func set(k string, v []byte) error {
+	getStorage().Call("setItem", js.ValueOf(k), js.ValueOf(string(v)))
+	// TODO(al) read back and check
```
mhutchinson, line 80: Is this a best practice thing, or literally something you observed while playing with this?

serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+	if exists(k) {
+		return os.ErrExist
+	}
+	getStorage().Call("setItem", js.ValueOf(k), js.ValueOf(string(v)))
+	// TODO(al) read back and check
+	return nil
+}
+
+// Load returns a Storage instance initialised from webstorage prefixed at root.
+func Load(root string, checkpoint *log.Checkpoint) (*Storage, error) {
```
mhutchinson, line 94: Passing the checkpoint into this method feels odd to me. It should be stored in the same root filesystem. Asking the client of this method to do the dance to call `cp, _ := ReadCheckpoint(foo); s, _ := Load(foo, cp)` seems risky and unclear as an API.

mhutchinson: Perhaps a method like `Exists(root)` or `Verify(root)` would be better. The client flow would then be:

```
var s Storage
if !webstorage.Exists(root) {
  s = webstorage.Create(root)
} else {
  s = webstorage.Load(root)
}
```

The implementation of Exists could just check there is a checkpoint there.

serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+	pk := []string{}
+	s := getStorage()
+	prefix := filepath.Join(fs.root, "leaves", "pending")
+	for i := 0; i < s.Get("length").Int(); i++ {
+		k := s.Call("key", js.ValueOf(i))
+		if js.Null().Equal(k) {
+			return nil, fmt.Errorf("key(%d) failed", i)
+		}
+		ks := k.String()
+		if strings.HasPrefix(ks, prefix) {
```
mhutchinson, line 133: It was only at this point that I realized there isn't even a pretense in the `s // getStorage` variable of a filesystem. It seems to be a collection of key / values, where the keys are structured such that by checking the prefix you can emulate a filesystem. I think that would be worth outlining at a high level in some documentation up there ^.

serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// Pending returns a pending leaf stored under PendingKey.
+func (fs *Storage) Pending(f string) ([]byte, error) {
+	return get(f)
+}
+
+// DeletePending removes a pending leaf stored under PendingKey.
+func (fs *Storage) DeletePending(f string) {
+	getStorage().Call("removeItem", f)
```
mhutchinson, line 147: Maybe check that `f` has the pending leaves prefix? And in `Pending`, above.

serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+	return &tile, nil
+}
+
+// StoreTile writes a tile out to disk.
+// Fully populated tiles are stored at the path corresponding to the level &
+// index parameters, partially populated (i.e. right-hand edge) tiles are
+// stored with a .xx suffix where xx is the number of "tile leaves" in hex.
+func (fs *Storage) StoreTile(level, index uint64, tile *api.Tile) error {
+	tileSize := uint64(tile.NumLeaves)
+	glog.V(2).Infof("StoreTile: level %d index %x ts: %x", level, index, tileSize)
```
mhutchinson, line 261: This is the only instance of `glog` in here. Would not using it allow a smaller wasm to be generated (faster) ?

If there was other critical logging then I'd not suggest this, but I don't see this logging ever being used as its V(2), but that can only be determined at runtime.

serverless/internal/storage/webstorage/fs.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// WriteCheckpoint stores a raw log checkpoint on disk.
+func (fs Storage) WriteCheckpoint(newCPRaw []byte) error {
+	oPath := filepath.Join(fs.root, layout.CheckpointPath)
+	return set(oPath, newCPRaw)
+}
+
+// ReadCheckpoint reads and returns the contents of the log checkpoint file.
+func ReadCheckpoint(root string) ([]byte, error) {
```
mhutchinson, line 287: Can we make this internal, and access this via `Storage.Checkpoint()` ? This is the same suggestion as that on `Load`, but a cleanup afterwards.

serverless/wasm/README.md
```diff
@@ -0,0 +1,29 @@
+# Serverless WASM
+
+This directory contains a version of the `serverless` tooling cross-compiled to WebAssembly (WASM).
+
+This is an entirely experimental thing, mostly intended to explore the possibility
```
mhutchinson, line 5: "entirely experimental" is strong! Consider putting `experimental` in the package name, as we did with the map.

serverless/wasm/README.md
```diff
@@ -0,0 +1,29 @@
...
+$ GOOS=js GOARCH=wasm go build -o serverless/wasm/main.wasm -tags wasm ./serverless/wasm/
+```
+
+then copy over the Go WASM shim into this directory:
+
+```bash
+$ cp "$(go env GOROOT)/misc/wasm/wasm_exec.js" serverless/wasm/
+```
+
+Run up a webserver to serve the files (use something which can infer the correct ContentType header to send), can use `goexec` for this (install `goexec` with `go get -u github.com/shurcooL/goexec`):
```
mhutchinson, line 23:
```suggestion
Run up a webserver to serve the files (use something which can infer the correct ContentType header to send). You can use `goexec` for this (install `goexec` with `go get -u github.com/shurcooL/goexec`):
```

serverless/wasm/main.go
```diff
@@ -0,0 +1,225 @@
...
+// Build with:
+//   GOOS=js GOARCH=wasm go build -o serverless/wasm/main.wasm -tags wasm ./serverless/wasm/
+//
+// To view locally:
+//   # Copy Go wasm JS shim:
+//   cp "$(go env GOROOT)/misc/wasm/wasm_exec.js" serverless/wasm/
+//   # Run a webserver serving the wasm dir
+//   # e.g. using goexec (to install goexec: go get -u github.com/shurcooL/goexec):
+//   goexec 'http.ListenAndServe(`:8080`, http.FileServer(http.Dir(`serverless/wasm`)))'
+// Then point your browser at the server.
```
mhutchinson, line 29: Consider removing from here and replacing with a pointer to the README. This is asking to go out of sync.

serverless/wasm/main.go
```diff
@@ -0,0 +1,225 @@
...
+
+const (
+	logPrefix = "log"
+	ecosystem = "WASM Log v0"
+)
+
+var (
+	logStorage *webstorage.Storage
+)
+
```
mhutchinson, line 59: Just to note I haven't really reviewed this last file past this point. A quick scan shows nothing that freaks me out, but I think there's mileage in addressing the other comments first to make sure my mental model of this is even in the right ballpark. Reviewing the below for anything detailed requires a context switch into application / user / xss-security type of thinking which I need a rest before doing :-)

serverless/internal/storage/webstorage/fs.go
AlCutter: This is because I have no idea how these calls are supposed to let the Go caller know when if it's failed to do its thing.
In the JS world it should raise an exception but I can't find any docs on what that looks like from Go's PoV...
`Get()` will at least return a `Null` if the thing I asked for isn't there.

serverless/internal/storage/webstorage/fs.go
AlCutter: This is the same as the `storage/fs` package it was forked from.

The reason for this is because the storage should not have to know about signatures/verifiers - all of that is handled at the caller's level (both opening and passing in here, as well as signing and persisting new CPs for larger trees).

serverless/internal/storage/webstorage/fs.go
AlCutter: JavaScript is the wrong thing I think (because Node, etc.), but I updated to call out browsers and the reason why though.

serverless/internal/storage/webstorage/fs.go
AlCutter: Yup, that's right - it is a key value store.
I don't really want to duplicate the documentation of the WebStorage API, but I've added a link to it in the comment about browsers at the top.

serverless/internal/storage/webstorage/fs.go
AlCutter: done

serverless/internal/storage/webstorage/fs.go
AlCutter: Only saves ~700 bytes in the complied file, but you're right to call it out - removed it.

serverless/internal/storage/webstorage/fs.go
AlCutter: Sadly not, for the same reason.

serverless/wasm/README.md
AlCutter: Probably a good idea - done

serverless/wasm/README.md
AlCutter: done

serverless/wasm/main.go
AlCutter: good call, done

serverless/internal/storage/webstorage/fs.go
mhutchinson: I suggested that based on the import of `syscall/js` but the new docs look good.

serverless/experimental/wasm/main.go
```diff
@@ -0,0 +1,219 @@
...
+			panic(err)
+		}
+		logStorage, err = webstorage.Load(logPrefix, cp)
+		if err != nil {
+			panic(err)
+		}
+	}
+	showCP()
+	go monitor()
+	<-make(chan bool)
```
mhutchinson, line 218: This is odd. Is this a wasm idiom?

mhutchinson (approved):

serverless/experimental/wasm/main.go
AlCutter: I guess it's not unlike the `http.ListenAndServe` thing - if you don't do this the program exits (in the WASM case that's bad since the JS world can call into the funcs it's published)

PR #415, opened 2021-07-21
Combine result and error channels together

mhutchinson:
The Bulk method now takes ownership of the channel too. This simplifies control flow.

This came up as part of another piece of work and has been separated into its own PR.


AlCutter:
Lint causing the presubmit to fail: `clone/internal/download/batch_test.go:42:2: S1000: should use for range instead of for { select {} }`

AlCutter (approved):

PR #418, opened 2021-07-22
[Serverless] Instructions & pretty consoles

AlCutter:
- Adds simple user instructions at the top of the page
- Just for fun, make the consoles look a little bit like VT323 terms


AlCutter:
Top of stack now

AlCutter:
These got merged via #420

PR #417, opened 2021-07-22
[Serverless] Add checkpoint sigs to the wasm experiment

AlCutter:
- This allows us to use the `client` package
- Use the `client` package in the monitor to do consistency checking of CPs

PR #420, opened 2021-07-22
[Serverless] Make WASM monitor inspect log entries

AlCutter:
- Print out log entries in the monitor console
- Use compact range to verify consistency and inclusion

(currently stacked on #417 and #418, will rebase)

### DIFF
--- a/serverless/experimental/wasm/main.go
+++ b/serverless/experimental/wasm/main.go
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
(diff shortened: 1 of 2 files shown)


serverless/experimental/wasm/main.go
```diff
@@ -209,29 +236,39 @@ func monitor(ctx context.Context, f client.Fetcher) {
...
-			}
-			proof, err := pb.ConsistencyProof(ctx, cpCur.Size, cp.Size)
+		rCP := rf.NewEmptyRange(cpCur.Size)
+		for i := cpCur.Size; i < cp.Size; i++ {
+			l, err := client.GetLeaf(ctx, f, i)
 			if err != nil {
-				monMsg(err)
-				continue
+				monMsg(fmt.Sprintf("Failed to fetch leaf at %d: %v", i, err))
+				break monitorLoop
```
mhutchinson, line 246: `return` would be more idiomatic and remove the need for the label?

mhutchinson (approved):

serverless/experimental/wasm/main.go
AlCutter: That would kill the monitor though?
The break takes us back up to sleep and try again

PR #423, opened 2021-07-23
Add WASM serverless binary so we can run from github directly

AlCutter:


mhutchinson (commented):
Can we add a presubmit check that this hasn't changed? It's at least a step towards some kind of binary transparency :-)

AlCutter:
> Can we add a presubmit check that this hasn't changed? It's at least a step towards some kind of binary transparency :-)

Maybe, but it might be a little brittle - everyone will have to build with at least `--trimpath` and I imagine it'll break between golang revisions (and dependabot bumps etc.)

Currently, I'm not even sure this serve from GH will even work :)

mhutchinson (approved):
If it doesn't host well from github then let's roll this back.

PR #424, opened 2021-07-23
[Serverless] no checkpoint state in storage

AlCutter:
Storage was previously parsing a signed note as a checkpoint, now the caller of `log.Integrate` passes in the checkpoint data.

### DIFF
--- a/serverless/internal/storage/fs/fs.go
+++ b/serverless/internal/storage/fs/fs.go
@@ -24,7 +24,6 @@ import (
 	"strconv"
 
 	"github.com/golang/glog"
-	"github.com/google/trillian-examples/formats/log"
 	"github.com/google/trillian-examples/serverless/api"
 	"github.com/google/trillian-examples/serverless/api/layout"
 	"github.com/google/trillian-examples/serverless/internal/storage"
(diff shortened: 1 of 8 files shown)


serverless/internal/storage/fs/fs.go
```diff
@@ -53,14 +53,12 @@ type Storage struct {
...
 	nextSeq uint64
-	// checkpoint is the latest known checkpoint of the log.
-	checkpoint log.Checkpoint
 }
 
 const leavesPendingPathFmt = "leaves/pending/%0x"
 
 // Load returns a Storage instance initialised from the filesystem.
-func Load(rootDir string, checkpoint *log.Checkpoint) (*Storage, error) {
+func Load(rootDir string, checkpoint log.Checkpoint) (*Storage, error) {
```
mhutchinson, line 61: This seems a bit messy to copy the struct in to only check the size. It's also the only reference left in this file to the `log` package, so maybe just pass in the uint64 directly?

mhutchinson (approved):

serverless/internal/storage/fs/fs.go
AlCutter: Yeah, I _um_'d and _ah_'d about that too.
I've changed it, ta.

PR #427, opened 2021-07-28
Add a check that the log string doesn't change after TOFU

mhutchinson:
This string shouldn't change once the log is configured. If the string changes (e.g. new version) then this is effectively a new log.


mhutchinson:
Rebased on top of #428 and added test for the ecosystem change.

AlCutter (approved):

PR #433, opened 2021-08-02
Breaking change: rename column in DB

mhutchinson:
Newer versions of MariaDB use `offset` as a reserved word. Renamed to `toffset` to sidestep the problem.

This is a breaking change for users upgrading from a previous version; the new code will fail to read this column. The solution is to manually rename the column in the DB to `toffset`, or to recreate the DB and mirror from a clean slate.


smeiklej (approved):
LGTM!

PR #434, opened 2021-08-02
Fix Docker build

mhutchinson:
This got broken when `witness` directory was added at the top level


smeiklej (approved):
Looks good to me (but of course I know nothing about docker :), assuming the cloud build works!

PR #436, opened 2021-08-05
SumDB feeder for the generic witness

mhutchinson:
This constructs compact ranges from the SumDB API and feeds these to the witness.

Setting this up with docker is fairly simple and is documented in the README.

### DIFF
--- a/sumdbaudit/witness/README.md
+++ b/sumdbaudit/witness/README.md
@@ -1,0 +1,38 @@
+# Lightweight Witness for SumDB
+
+This directory contains a dockerized deployment for a lightweight witness and feeder.
+Unlike the SumDB mirror & witness implementation in this repository, this version only acquires and stores a minimal amount of information.
+This makes it much easier & cheaper to deploy.
+
+The feeder will poll the SumDB periodically, and if the size of its checkpoint is larger than the one committed to by the witness, then a consistency proof will be generated and sent to the witness.
+The witness is simply a deployment of the [generic witness](../../witness/golang/README.md).
+
+## Running
+
+### Initial Setup
+
+Before running the witness, first prepare a config directory containing `witness.config`.
+If this is copied anywhere other than `/etc/witness/config/witness.config` then `WITNESS_CONFIG_DIR` and `WITNESS_CONFIG_FILE` environment variables must be set when running docker compose later.
+
+You should also generate a keypair for the witness using `note.GenerateKey`, e.g. https://play.golang.org/p/uWUKLNK6h9v.
+
+### Running The Witness
+
+Change the environment variables below to the values you have generated, and set the config directory/file variables similarly if needed.
+
+```bash
+export WITNESS_PRIVATE_KEY="PRIVATE+KEY+WitnessName+privatekey"; \
+export WITNESS_PUBLIC_KEY="WitnessName+publickey"; \
+docker compose -f ./sumdbaudit/witness/docker-compose.yaml up -d
+```
+
+To confirm this is working you can run `curl localhost:8100/witness/v0/logs/sumdb/checkpoint` which should give something like:
+
+```
+go.sum database tree
+6353345
+d5kp6/RTbWj/e9AMga1hdcHZHcLdFKA2fjHfKJ8FfR4=
+
+— sum.golang.org Az3grhfKc0hf9eAH1x5p0VjY99pEe8l9JKKLGyf9F0m4JTrJjcsr9rUDh6kvNIl7vdzWqULpk3+azvpfJo9aOMZaYQE=
+— mdh.SumDB.Witness MwRoUvap0myxmJQ+D1EuU61mmID6Fu1anufFrU0E0FgqaVj8ZAjWSJ7eWzi8tQ4dOljZ2cVlDmSesoaeBAMC1t94Mgc
+```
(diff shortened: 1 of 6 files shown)


sumdbaudit/witness/cmd/feeder/main.go
```diff
@@ -0,0 +1,295 @@
...
+// feedOnce gets the latest checkpoint from the SumDB server, and if this is more
+// recent than the witness state then it will construct a compact range proof by
+// requesting the minimal set of tiles, and then update the witness with the new
+// checkpoint. Finally, it will update the feeder's view of the witness state.
+func (f *feeder) feedOnce(ctx context.Context) error {
+	sdbcp, err := f.sdb.LatestCheckpoint()
+	if err != nil {
+		return fmt.Errorf("failed to get latest checkpoint: %v", err)
+	}
+	if int64(f.wcp.Size) >= sdbcp.N {
```
smeiklej, line 131: If the witness' checkpoint is strictly bigger than the latest sumDB checkpoint then something has really gone wrong, right?  Is it worth treating this case separately and exiting/logging some kind of sadness?

sumdbaudit/witness/cmd/feeder/main.go
```diff
@@ -0,0 +1,295 @@
...
+		proof = append(proof, h)
+	}
+
+	wcpRaw, err := f.w.Update(ctx, *logID, sdbcp.Raw, proof)
+
+	if err != nil && !errors.Is(err, http.ErrCheckpointTooOld) {
+		return fmt.Errorf("failed to update checkpoint: %v", err)
+	}
+
+	// An optimization would be to immediately retry the Update if we get
```
smeiklej, line 158: Could you put this as an explicit TODO?  In particular if this is run in a multi-feeder scenario then it seems very possible that the feeder will fail quite frequently (potentially always?) if it waits until the next new checkpoint to retry.

smeiklej (commented):
Some very minor comments!  The biggest one is maybe just teasing apart two cases (equal to and strictly greater than).

sumdbaudit/witness/README.md
```diff
@@ -0,0 +1,38 @@
+# Lightweight Witness for SumDB
+
+This directory contains a dockerized deployment for a lightweight witness and feeder.
+Unlike the SumDB mirror & witness implementation in this repository, this version only acquires and stores a minimal amount of information.
+This makes it much easier & cheaper to deploy.
+
+The feeder will poll the SumDB periodically, and if the size of its checkpoint is larger than the one committed to by the witness, then a consistency proof will be generated and sent to the witness.
```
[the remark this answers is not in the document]
mhutchinson: Sure, done.

sumdbaudit/witness/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Change the environment variables below to the values you have generated, and set the config directory/file variables similarly if needed.
+
+```bash
+export WITNESS_PRIVATE_KEY="PRIVATE+KEY+WitnessName+privatekey"; \
+export WITNESS_PUBLIC_KEY="WitnessName+publickey"; \
+docker compose -f ./sumdbaudit/witness/docker-compose.yaml up -d
+```
+
+To confirm this is working you can run `curl localhost:8100/witness/v0/logs/sumdb/checkpoint` which should give something like:
```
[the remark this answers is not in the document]
mhutchinson: Yep, but then it almost immediately gets updated by the feeder that is also started in the `docker-compose`. Try this locally; you should find the instructions here aren't a lie unless you're super fast!

sumdbaudit/witness/cmd/feeder/main.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+}
+
+// feeder encapsulates the main logic and state of the feeder.
+// The residual logic outside of this should simply be initialization and
+// orchestration with a timer.
+// Note that feeder maintains its own state that represents the witness state.
+// This is optimized for the case where this is the only feeder for this log to
+// the witness. If the witness state is updated outwith this feeder, then the
```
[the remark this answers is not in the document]
mhutchinson: Achievement Unlocked, I guess? :-)

Changed to avoid this word for the international community though.

sumdbaudit/witness/cmd/feeder/main.go
mhutchinson: In distributed systems it could happen that a "server" (actually a collection of server instances) is not internally consistent and thus making 2 quick requests gets you a fresh answer and a stale answer. Eventually they will become consistent, and nothing nefarious is happening here.

If anything nefarious is happening, then it will eventually get caught out by the consistency check when the tree next grows.

I could separate this out and Warn in the case of a size regression as a curio, but I wouldn't make it bail at this point. If you're still attached to the printing of a stronger message then comment here and I'll make it so.

sumdbaudit/witness/cmd/feeder/main.go
mhutchinson: I initially had this as an explicit TODO but decided not to for 2 reasons:
 * I currently have no intention of doing this
 * The primary intention for this feeder is to feed a local witness; this is documented on the `feeder` above (This is optimized for the case where this is the only feeder for this log to the witness.)

I leave this message here as a clear signpost for anyone that fancies changing that, but I'm actively working on reducing the number of TODOs to only those we would actually do given a bit of spare time. The trigger for doing this wouldn't be spare time, but the intention to run this as a collection of feeders to a single witness.

Leaving the comment open; resolve this if you're convinced, or holler back if you're not and I'll cancel my TODO-reduction thread and just make it so.

smeiklej (approved):
LGTM!  Thanks for the clarifications.

PR #437, opened 2021-08-05
[witness] the missing bananas secret key emerges

smeiklej:


mhutchinson (approved):
Maybe revert the changes to go.mod. The changes to the test shouldn't be changing that file. If we need to do go mod tidy, let's do it in another PR.

PR #440, opened 2021-08-11
A note parser and verifier for checkpoints 

mhutchinson:
This handles parsing the note, and verifying that the log has signed it. Optionally, other required verifiers can be specified. If these are provided, then all of the signatures from these verifiers must be present.

This also checks the first line of the message is the LogID that was set to be expected when the parser was constructed. This part isn't decided (other proposals exist for where the LogID should live) but I've added this check here as a strawman.

### DIFF
--- a/formats/log/note.go
+++ b/formats/log/note.go
@@ -15,70 +15,43 @@
 package log
 
 import (
+	"errors"
 	"fmt"
 
 	"golang.org/x/mod/sumdb/note"
 )
 
-// CheckpointParser is responsible for parsing checkpoints for a specific log.
-// Any valid note must have been signed by all verifiers known by this parser.
-type CheckpointParser struct {
-	logVerifier       note.Verifier
-	logID             string
-	otherVerifiers    note.Verifiers
-	requiredOtherSigs int
-}
+// ParseCheckpoint takes a raw checkpoint as bytes and returns a parsed checkpoint
+// providing that a valid log signature is found, the checkpoint unmarshals
+// correctly, and the log origin is that expected. In all other cases, an empty
+// checkpoint is returned. The underlying note is always returned where possible.
+// The signatures on the note will include the log signature if no error is returned,
+// plus any signatures from otherVerifiers that were found.
+func ParseCheckpoint(origin string, logVerifier note.Verifier, otherVerifiers []note.Verifier, chkpt []byte) (Checkpoint, *note.Note, error) {
+	vs := make([]note.Verifier, len(otherVerifiers)+1)
+	vs[0] = logVerifier
+	if n := copy(vs[1:], otherVerifiers); n < len(otherVerifiers) {
+		return Checkpoint{}, nil, errors.New("failed to create verifier list")
+	}
+	verifiers := note.VerifierList(vs...)
 
-// NewCheckpointParser creates a parser for checkpoints that must be signed by the log
-// which is verified by logVKey, and *all* of the other verifiers under otherVerifiers.
-func NewCheckpointParser(logVKey string, logID string, otherVerifiers ...string) (CheckpointParser, error) {
-	logVerifier, err := note.NewVerifier(string(logVKey))
+	n, err := note.Open(chkpt, verifiers)
 	if err != nil {
-		return CheckpointParser{}, fmt.Errorf("NewVerifier(%s): %v", logVKey, err)
+		return Checkpoint{}, n, fmt.Errorf("failed to verify signatures on checkpoint: %v", err)
 	}
 
-	ovs := make([]note.Verifier, len(otherVerifiers))
-	for i, k := range otherVerifiers {
-		ovs[i], err = note.NewVerifier(k)
-		if err != nil {
-			return CheckpointParser{}, fmt.Errorf("NewVerifier(%s): %v", k, err)
+	for _, s := range n.Sigs {
+		if s.Hash == logVerifier.KeyHash() && s.Name == logVerifier.Name() {
+			// The log has signed this checkpoint. It is now safe to parse.
+			cp := &Checkpoint{}
+			if _, err := cp.Unmarshal([]byte(n.Text)); err != nil {
+				return Checkpoint{}, n, fmt.Errorf("failed to unmarshal checkpoint: %v", err)
+			}
+			if cp.Origin != origin {
+				return Checkpoint{}, n, fmt.Errorf("got Origin %q but expected %q", cp.Origin, origin)
+			}
+			return *cp, n, nil
 		}
 	}
-	return CheckpointParser{
-		logVerifier:       logVerifier,
-		logID:             logID,
-		otherVerifiers:    note.VerifierList(ovs...),
-		requiredOtherSigs: len(otherVerifiers),
-	}, nil
-}
-
-// Parse returns a checkpoint parsed from the raw note, and any error if the note could
-// not be parsed, or any of the required signatures were missing.
-func (p CheckpointParser) Parse(chkptRaw []byte) (Checkpoint, *note.Note, error) {
-	r := &Checkpoint{}
-	n, err := note.Open(chkptRaw, note.VerifierList(p.logVerifier))
-	if err != nil {
-		return *r, n, fmt.Errorf("failed to verify log signature on checkpoint: %v", err)
-	}
-	if _, err := r.Unmarshal([]byte(n.Text)); err != nil {
-		return *r, n, fmt.Errorf("failed to unmarshal new checkpoint: %v", err)
-	}
-	// TODO(mhutchinson): This assumes the first line is the LogID.
-	// This is proposal 2B. If we put the LogID further in the message (e.g. 4th line; 1A)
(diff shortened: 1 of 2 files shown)


formats/log/note.go
```diff
@@ -0,0 +1,84 @@
...
+	// then this should be updated.
+	if r.Ecosystem != p.logID {
+		return *r, n, fmt.Errorf("got logID %q but expected %q", r.Ecosystem, p.logID)
+	}
+
+	if p.requiredOtherSigs == 0 {
+		return *r, n, nil
+	}
+
+	if n, err = note.Open(chkptRaw, p.otherVerifiers); err != nil {
```
smeiklej, line 77: I can see the reasoning behind this but it seems awkward to have two calls to `note.Open`, especially when the second one comes after you've already parsed the note.  If you did one call to `note.Open` with both the log and other verifiers then you could also remove lines 73-75, but on the other hand I guess you'd have no way to distinguish between a log or other signature failing which would make the error messaging less helpful.  So maybe I'm just saying I wish Note made it easier to find out which signatures verified and there is no actual suggestion here :).

smeiklej (approved):
I guess this is still a draft but I'm approving it anyway!

formats/log/note.go
mhutchinson: I realize it looks more awkward in here, but if it puts our minds at rest, the performance is pretty identical (if anything, this way seems faster, which is odd):

Benchmark with only one `note.Open` (https://github.com/mhutchinson/trillian-examples/pull/new/checkpointParserSimpler):
```
BenchmarkParse/bare_minimum-4                 7752        174560 ns/op        1024 B/op          14 allocs/op
BenchmarkParse/one_verifier-4                 4048        395803 ns/op        1520 B/op          18 allocs/op
BenchmarkParse/many_verifiers-4               1612        674900 ns/op        2450 B/op          25 allocs/op
BenchmarkParse/many_sigs,_1_verifier-4      4126        347926 ns/op        3953 B/op          41 allocs/op
```

Benchmark from this code with two `note.Open`:
```
BenchmarkParse/bare_minimum-4                 8337        144167 ns/op        1504 B/op          17 allocs/op
BenchmarkParse/one_verifier-4                 3813        304662 ns/op        3056 B/op          35 allocs/op
BenchmarkParse/many_verifiers-4               2007        578182 ns/op        5154 B/op          55 allocs/op
BenchmarkParse/many_sigs,_1_verifier-4      4092        298907 ns/op        8017 B/op          81 allocs/op
```

formats/log/note.go
```diff
@@ -0,0 +1,84 @@
...
+// Any valid note must have been signed by all verifiers known by this parser.
+type CheckpointParser struct {
+	logVerifier       note.Verifier
+	logID             string
+	otherVerifiers    note.Verifiers
+	requiredOtherSigs int
+}
+
+// NewCheckpointParser creates a parser for checkpoints that must be signed by the log
+// which is verified by logVKey, and *all* of the other verifiers under otherVerifiers.
```
AlCutter, line 33: The `all` other verifiers is surprising, I'd expected to see something like "MUST be signed by the log, MAY be cosigned by a subset of these witnesses" and either leave the N of M check to the caller of `Parse` (via the returned `Note.Sigs`) or pass in the N to that call?

formats/log/note.go
```diff
@@ -0,0 +1,84 @@
...
+func NewCheckpointParser(logVKey string, logID string, otherVerifiers ...string) (CheckpointParser, error) {
```
AlCutter, line 34: I'd be tempted to pass in `note.Verifier` here rather than the string so that folks can use their own implementations (e.g. `tink` based, or implementations using algos which aren't present in the `note.NewVerifier` func)

formats/log/note.go
```diff
@@ -0,0 +1,84 @@
...
+		logVerifier:       logVerifier,
+		logID:             logID,
+		otherVerifiers:    note.VerifierList(ovs...),
+		requiredOtherSigs: len(otherVerifiers),
+	}, nil
+}
+
+// Parse returns a checkpoint parsed from the raw note, and any error if the note could
+// not be parsed, or any of the required signatures were missing.
+func (p CheckpointParser) Parse(chkptRaw []byte) (Checkpoint, *note.Note, error) {
```
AlCutter, line 57: If I convinced you about the verifiers above, then maybe this could become `log.ParseCheckpoint(cpRaw, logID, logSigV, otherSigV, numOtherSigsRequired) (Checkpoint, *note.Note, error)`?

formats/log/note.go
AlCutter: I think there's a slight functional difference with the current impl in that the returned `n` will appear not to have had a valid sig from the log in its `Sigs` array (since we're overwriting the `n` which did have that record).

It's implicit that it did, of course, but maybe a bit weird that it's not present?

mhutchinson: I've changed this behaviour now. There's only one call to note.Open and all matching signatures are returned.

formats/log/note.go
mhutchinson: I've met you half way. The number of signatures required is left to the caller to verify.

formats/log/note.go
mhutchinson: Take a look at the new API and see what you think.

formats/log/note.go
mhutchinson: Done.

formats/log/note.go
```diff
@@ -0,0 +1,57 @@
...
+	"golang.org/x/mod/sumdb/note"
+)
+
+// ParseCheckpoint takes a raw checkpoint as bytes and returns a parsed checkpoint
+// providing that a valid log signature is found, the checkpoint unmarshals
+// correctly, and the log origin is that expected. In all other cases, an empty
+// checkpoint is returned. The underlying note is always returned where possible.
+// The signatures on the note will include the log signature if no error is returned,
+// plus any signatures from otherVerifiers that were found.
+func ParseCheckpoint(origin string, logVerifier note.Verifier, otherVerifiers []note.Verifier, chkpt []byte) (Checkpoint, *note.Note, error) {
```
AlCutter, line 30: Yeah, nice - I think this has become (even) more useful with the recent `origin` discussion and change.

formats/log/note.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"golang.org/x/mod/sumdb/note"
+)
+
+// ParseCheckpoint takes a raw checkpoint as bytes and returns a parsed checkpoint
+// providing that a valid log signature is found, the checkpoint unmarshals
```
AlCutter, line 25: Consider breaking this list out into a MD list? Or if that's too much, introduce the list with a colon - I had a parse exception reading through this :)

formats/log/note.go
```diff
@@ -0,0 +1,57 @@
...
+// correctly, and the log origin is that expected. In all other cases, an empty
+// checkpoint is returned. The underlying note is always returned where possible.
```
AlCutter, line 27: Any reason not to return `*Checkpoint` (nil on error)?

formats/log/note.go
```diff
@@ -0,0 +1,57 @@
...
+// The signatures on the note will include the log signature if no error is returned,
+// plus any signatures from otherVerifiers that were found.
+func ParseCheckpoint(origin string, logVerifier note.Verifier, otherVerifiers []note.Verifier, chkpt []byte) (Checkpoint, *note.Note, error) {
+	vs := make([]note.Verifier, len(otherVerifiers)+1)
```
AlCutter, line 31: `vs := append([]note.Verifier{logVerifier}, otherVerifiers...)` is a more concise way of doing these 5 lines.

If you're worried about the prealloc optimisation you could even `vs := append(append(make([]note.Verifier, 0, len(...), logVerifier), otherVerifiers...)`

formats/log/note.go
mhutchinson: Yeah, the addition of that extra check is what convinced me that getting all clients to perform this extra check was going to lead to bugs!

formats/log/note.go
mhutchinson: Formatted

formats/log/note.go
mhutchinson: Because early code reviews on my code written in Go led me to believe that pointers shouldn't be returned unless there's a good reason for it. I don't know who to believe any more. I suspect this is one of those things where everyone else has an opinion, and as someone who just wants an easy life here I'm the one that really loses out :-)

nil would be more natural to me, so I'll change it. But you better have my back when the avoid-pointers brigade turn up.

formats/log/note.go
mhutchinson: Done.

formats/log/note.go
AlCutter: No worries, I'll just leave you with this bit of paper telling you where I'll be if you to find need me later on... :dash:

AlCutter (approved):

PR #441, opened 2021-08-11
Speed up serverless tests

mhutchinson:
Running the integration tests in parallel saves a good minute on the total test time for `go test ./...`.

The change to the marshal test to only check roundtripping for a few interesting tile sizes drops the test time from 8+s to under 1s.


smeiklej (approved):
Looks great!

PR #447, opened 2021-08-18
Extended documentation to discuss roles

mhutchinson:


witness/README.md
```diff
@@ -1,15 +1,45 @@
...
-misbehavior on the part of logs can be prevented or detected. This is done by
-ensuring that the _checkpoints_ published by logs are _consistent_, which means 
-the log is presenting the same version of its contents to everyone and
-thus isn't carrying out a _split-view attack_.
+## Overview
 
-In terms of checkpoints, we consider witness implementations that use the
+Witnessing is an important part of log ecosystems that prevents logs from
+successfully accomplishing a _split-view attack_. A split-view attack is
+one in which the log prevents different populations with different views.
```
AlCutter, line 7: `prevents` -> `presents`

witness/README.md
```diff
@@ -1,15 +1,45 @@
...
+As a counterpoint, a common situation is to have the actor playing the witness role
+also polling the log for checkpoints and proofs (i.e. playing the feeder role), and
+making cosigned checkpoints available via some API (distributor).
+
+## Generic Witness Implementation
+
+This repository provides a witness that will work for any log that uses the
 [format defined in this
 repository](https://github.com/google/trillian-examples/tree/master/formats/log).
+Each log will require a custom feeder implementation to provide the binding
```
AlCutter, line 38: It's probably less than "each log" right?
Perhaps more along the lines of "type of log" or something (although I'm not sure exactly what the right phrasing is)

witness/README.md
```diff
@@ -46,3 +75,8 @@ defined above.  Currently it contains implementations in the following languages
   This can be deployed as a smart contract on the Ethereum blockchain.  There are 
   two main contracts, each of which supports one of the consistency checks 
   described above.
+
+## Feeders
+
+* [Go SumDB](https://github.com/google/trillian-examples/tree/master/sumdbaudit/witness)
+* [Serverless Logs](https://github.com/google/trillian-examples/tree/master/serverless/cmd/feeder)
```
AlCutter, line 82: nit: `Logs` -> `logs` for case consistency with the rest of this page.

witness/README.md
```diff
@@ -1,15 +1,45 @@
...
-ensuring that the _checkpoints_ published by logs are _consistent_, which means 
-the log is presenting the same version of its contents to everyone and
-thus isn't carrying out a _split-view attack_.
+## Overview
 
-In terms of checkpoints, we consider witness implementations that use the
+Witnessing is an important part of log ecosystems that prevents logs from
+successfully accomplishing a _split-view attack_. A split-view attack is
+one in which the log prevents different populations with different views.
+Each of these views is append-only, and thus all clients checking consistency
```
smeiklej, line 8: I would say that "Each of these views *may be* append-only" and then "all clients [...] *would be* satisfied", since a (somewhat dumb) log could even provide inconsistent checkpoints to the same population (like for example one that it knows doesn't check consistency proofs)?

witness/README.md
```diff
@@ -1,15 +1,45 @@
...
+proofs are satisfied, yet the log is not fulfilling its claim of providing a
+_globally consistent_ list.
+
+For each witnessed log, a Witness will cosign log _checkpoints_ only if it
```
smeiklej, line 12: Very much a nit: Here and several other places, you use the simple future tense (the witness will do this, each log will do that), which seems unnecessary.  How about just the present? 😊

witness/README.md
```diff
@@ -1,15 +1,45 @@
...
+
+We consider witnessing to be composed of three primary _roles_:
+  1. Feeder: acquires checkpoints and consistency proofs, and provides these
+     to the witness
+  2. Witness: maintains and cosigns a checkpoint for each log, and updates
+     this state to a new checkpoint only when convinced that the delta is only
+     appending new leaves
+  3. Distributor: makes cosigned checkpoints available to clients
+
+Note that this does not mean there needs to be 3 separate actors playing these roles.
```
smeiklej, line 28: Another nitty nit: three instead of 3? (https://www.grammarly.com/blog/when-to-spell-out-numbers/)

witness/README.md
```diff
@@ -1,15 +1,45 @@
...
+For each witnessed log, a Witness will cosign log _checkpoints_ only if it
+can verify that it is consistent with the last cosigned checkpoint, i.e. that
+this new checkpoint represents only new values being appended to the log.
+Clients that acquire checkpoints signed by multiple independent witnesses can
+have more confidence that they are seeing a truly globally consistent log.
+
+## Roles
+
+We consider witnessing to be composed of three primary _roles_:
+  1. Feeder: acquires checkpoints and consistency proofs, and provides these
```
smeiklej, line 21: Saying that the feeder "acquires" these things makes it sound like they have to go out and get them, instead of (maybe) passively receiving them from the log or another entity.

witness/README.md
```diff
@@ -46,3 +75,8 @@ defined above.  Currently it contains implementations in the following languages
   This can be deployed as a smart contract on the Ethereum blockchain.  There are 
   two main contracts, each of which supports one of the consistency checks 
   described above.
+
+## Feeders
+
+* [Go SumDB](https://github.com/google/trillian-examples/tree/master/sumdbaudit/witness)
```
smeiklej, line 81: This is potentially confusing, since what's here is a combined feeder and witness.  Maybe just add a little comment pointing that out?

smeiklej (commented):
Some very minor comments!

witness/README.md
mhutchinson: I removed the word because I added explainer text to address another comment.

mhutchinson (commented):
Well that escalated quickly :-) PTAL. Will submit tomorrow if no more comments appear beforehand.

witness/README.md
```diff
@@ -46,3 +73,10 @@ defined above.  Currently it contains implementations in the following languages
   This can be deployed as a smart contract on the Ethereum blockchain.  There are 
   two main contracts, each of which supports one of the consistency checks 
   described above.
+
+## Feeders
+
+* [Go SumDB](https://github.com/google/trillian-examples/tree/master/sumdbaudit/witness) :
+    a custom feeder for sum.golang.org that can be easily deployed with this generic witness implementation
```
smeiklej, line 80: Last nit: this readme is for multiple witness implementations, so it's not clear here which one you mean. :)

smeiklej (approved):

witness/README.md
mhutchinson: Made more specific.

PR #450, opened 2021-08-19
Use locking in reporter to avoid anomolies

mhutchinson:
This was using a lock-free mechanism before, which worked but isn't a good practice.

This change was initially created to avoid seeing > 100% work in any reporting window. It seems to reduce that but not completely avoid it. The reason for that is that work isn't cleanly divided into reporting chunks and can overlap reporting periods, but only be counted in one of them.

### DIFF
--- a/clone/cmd/ctclone/ctclone.go
+++ b/clone/cmd/ctclone/ctclone.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


clone/cmd/ctclone/ctclone.go
```diff
@@ -177,27 +178,37 @@ type reporter struct {
...
+		workRatio := r.epochWorked.Seconds() / elapsed.Seconds()
 
-	remaining := r.treeSize - r.lastReported - 1
-	rate := float64(r.lastWorked-r.lastReported) / elapsed.Seconds()
-	eta := time.Duration(float64(remaining)/rate) * time.Second
-	glog.Infof("%.1f leaves/s, last leaf=%d (remaining: %d, ETA: %s), time working=%.1f%%", rate, r.lastReported, remaining, eta, 100*workRatio)
+		remaining := r.treeSize - r.lastReported - 1
+		rate := float64(r.lastWorked-r.lastReported) / elapsed.Seconds()
+		eta := time.Duration(float64(remaining)/rate) * time.Second
+		glog.Infof("%.1f leaves/s, last leaf=%d (remaining: %d, ETA: %s), time working=%.1f%%", rate, r.lastReported, remaining, eta, 100*workRatio)
```
AlCutter, line 195: Just a thought - you could make a snapshot copy of the values under lock and then release that so that you're not locked for the duration of the logging i/o on this line?

maybe something like:
```go
...

lw, lr, ew := func() (uint64, uint64, time.Duration) {
  r.mu.RLock()
  defer r.mu.RUnlock()
  return r.lastWorked, r.lastReported, r.epochWorked
}

// do calculations and log using local vars from above

// lock and update stats
}
```

AlCutter (approved):

clone/cmd/ctclone/ctclone.go
mhutchinson: Nice. I've taken this on, and reconsidered what the lock was actually protecting. Instead of 3 lock blocks, there are now only 2 which also ensures work can't get missed, which was possible in the previous implementation (epochWorked could be read, then incremented, then zeroed).

AlCutter (approved):

PR #452, opened 2021-08-19
Extracted core of bulk cloning into a library

mhutchinson:
This is the common code that any clone tool is likely to need. Specifically, this library is usable by the sumdbclone tool that I've got pending and significantly reduces a lot of duplication.

### DIFF
--- a/clone/internal/clone/clone.go
+++ b/clone/internal/clone/clone.go
@@ -101,7 +101,7 @@ func (c Cloner) Clone(ctx context.Context, treeSize uint64, fetcher download.Bat
 	bs := next
 	for br := range brc {
 		if br.Err != nil {
-			glog.Exit(err)
+			return fmt.Errorf("failed to get leaf at %d: %v", next, br.Err)
 		}
 		workDone := r.trackWork(next)
 		next++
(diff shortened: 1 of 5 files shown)


mhutchinson:
This will ideally be submitted after #454 which fixes a problem that came to light when testing this. If so, then one of the test cases can be re-enabled in this PR.

clone/internal/clone/clone.go
```diff
@@ -0,0 +1,193 @@
...
+type Cloner struct {
+	workers        uint
+	fetchBatchSize uint
+	writeBatchSize uint
+
+	db *logdb.Database
+}
+
+// NewCloner creates a new Cloner with the given performance parameters and database.
+func NewCloner(wokers, fetchBatchSize, writeBatchSize uint, db *logdb.Database) Cloner {
```
AlCutter, line 41: How about renaming this package `cloner` and calling this `New`?
(packages often seem to be nouns too)

AlCutter (approved):

clone/internal/clone/clone.go
mhutchinson: Done.

PR #453, opened 2021-08-19
SumDB clone tool that downloads the DB quickly 

mhutchinson:
This is consistent with the ctclone tool and uses much of the same machinery to get the job done. There is a little additional complexity here where we need to always be downloading a tile per worker to realize the parallelism dream. This means that when the previous download finished mid-tile, we need to do a little neatening up before we can invoke the Bulk command. Otherwise, this drops in cleanly if I say so myself.

### DIFF
--- a/clone/README.md
+++ b/clone/README.md
@@ -37,7 +37,7 @@ MariaDB [(none)]> FLUSH PRIVILEGES;
 ## Tuning
 
 The clone library logs information as it runs using `glog`.
-Providing `--alsologtostderr` is provided to any tool using the library, you should see output such as the following during the cloning process:
+Providing `--alsologtostderr` is passed to any tool using the library, you should see output such as the following during the cloning process:
 
 ```
 I0824 11:09:23.517796 2881011 clone.go:71] Fetching [4459168, 95054738): Remote leaves: 95054738. Local leaves: 4459168 (0 verified).
@@ -47,7 +47,7 @@ I0824 11:09:38.518542 2881011 clone.go:177] 1024.0 leaves/s, last leaf=4470430 (
 ```
 
 When tuning, it is recommended to also provide `--v=1` to see more verbose output.
-In particular, this will allow you to see if the tool is encountering errors (such as being told to back off) by the log server if you see lines such as `Retryable error getting data` in the output.
+In particular, this will allow you to see if the tool is encountering errors (such as being told to back off) by the log server, e.g. if you see lines such as `Retryable error getting data` in the output.
 
 Assuming you aren't being rate limited, then optimization goes as follows:
   1. Get the `working` percentage regularly around 100%: this measures how much time is being spent writing to the database. To do this, increase the number of `workers` to ensure that data is always available to the database writer
--- a/clone/README.md
+++ b/clone/README.md
@@ -1,15 +1,31 @@
 # Log Cloner
 
-A log client that copies a CT log to a local database for processing.
-The tool downloads batches of leaves in parallel, but always writes them to the local database in sequence.
-This ensures there are no missing ranges, which keeps state tracking easier.
+This directory contains a library, database, and tools for cloning transparency logs.
+The core library and database is log-agnostic, and each tool tailors this generic library to a specific log.
 
-One important implementation feature is that the download tools will exponentially back off if there are errors
-communicating with the log; this prevents the client from performing a DoS on any log it is downloading.
+The core library attempts to balance optimization of the following goals:
+  1. Downloading as quickly as possible
+  2. Backing off when requested by the log (i.e. not DoSing the log)
+  3. Simple local state / recovery
 
-## Setup
+This is achieved by:
+  1. Downloading batches of leaves in parallel
+  2. Using exponential backoff on transient failures
+  3. Writing leaves to the local database strictly in sequence
+     1. This ensures there are no missing ranges, which keeps state tracking easier
 
-In MariaDB, create a database and user:
+These tools are written to clone the log at a point in time.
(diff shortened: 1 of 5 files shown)


clone/README.md
```diff
@@ -18,40 +34,32 @@ MariaDB [(none)]> GRANT ALL PRIVILEGES ON google_xenon2022.* TO 'clonetool'@loca
...
-go run ./clone/cmd/ctclone --alsologtostderr --v=1 --log_url https://ct.googleapis.com/logs/xenon2022/ --mysql_uri 'clonetool:letmein@tcp(localhost)/google_xenon2022'
-```
+## Tuning
 
-See the optional flags in the `ctclone` tool to configure tuning parameters.
-## Docker
-
-To build a docker image, run the following from the `trillian-examples` root directory:
+The clone library logs information as it runs using `glog`.
+Providing `--alsologtostderr` is provided to any tool using the library, you should see output such as the following during the cloning process:
```
AlCutter, line 40: `is provided` -> `is passed` (repetition)

clone/README.md
```diff
@@ -18,40 +34,32 @@ MariaDB [(none)]> GRANT ALL PRIVILEGES ON google_xenon2022.* TO 'clonetool'@loca
...
-docker build . -t ctclone -f ./clone/cmd/ctclone/Dockerfile
+I0824 11:09:23.517796 2881011 clone.go:71] Fetching [4459168, 95054738): Remote leaves: 95054738. Local leaves: 4459168 (0 verified).
+I0824 11:09:28.519257 2881011 clone.go:177] 1202.8 leaves/s, last leaf=4459168 (remaining: 90595569, ETA: 20h55m21s), time working=24.2%
+I0824 11:09:33.518444 2881011 clone.go:177] 1049.6 leaves/s, last leaf=4465183 (remaining: 90589554, ETA: 23h58m29s), time working=23.0%
+I0824 11:09:38.518542 2881011 clone.go:177] 1024.0 leaves/s, last leaf=4470430 (remaining: 90584307, ETA: 24h34m21s), time working=23.3%
 ```
 
-This can be pointed at a local MySQL instance running outside of docker using:
+When tuning, it is recommended to also provide `--v=1` to see more verbose output.
+In particular, this will allow you to see if the tool is encountering errors (such as being told to back off) by the log server if you see lines such as `Retryable error getting data` in the output.
```
AlCutter, line 50: `by the log server if you see` -> ` e.g. if you see log lines such as...`?

clone/cmd/sumdbclone/sumdbclone.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/google/trillian-examples/clone/internal/verify"
+	"github.com/google/trillian-examples/clone/logdb"
+	sdbclient "github.com/google/trillian-examples/sumdbaudit/client"
+	"github.com/google/trillian/merkle/rfc6962"
+
+	_ "github.com/go-sql-driver/mysql"
+)
+
+var (
+	height         = flag.Int("h", 8, "tile height")
```
AlCutter, line 36: More verbose flag names and descriptions for these first two would probably be good for folks jumping in.

clone/cmd/sumdbclone/sumdbclone.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func main() {
+	flag.Parse()
+
+	if len(*mysqlURI) == 0 {
+		glog.Exit("Missing required parameter 'mysql_uri'")
+	}
+
+	ctx := context.Background()
```
AlCutter, line 50: nitty: could move this down to L63?

clone/README.md
mhutchinson: Ta.

clone/README.md
mhutchinson: Sure.

clone/cmd/sumdbclone/sumdbclone.go
mhutchinson: Made `height` a constant for now because realistically anyone changing this should know what they are doing. Added something for the vkey.

clone/cmd/sumdbclone/sumdbclone.go
mhutchinson: I prefer to determine the context up near the top of main functions. It's possible that the other init stuff would take it in the future, and then it's just going to move up. I don't think moving it down adds any clarity.

clone/cmd/sumdbclone/sumdbclone.go
```diff
@@ -33,13 +33,16 @@ import (
...
-	height         = flag.Int("h", 8, "tile height")
-	vkey           = flag.String("k", "sum.golang.org+033de0ae+Ac4zctda0e5eza+HJyk9SxEdh+s3Ux18htTTAD8OuAn8", "key")
+	vkey           = flag.String("vkey", "sum.golang.org+033de0ae+Ac4zctda0e5eza+HJyk9SxEdh+s3Ux18htTTAD8OuAn8", "The verification key for the log checkpoints")
 	mysqlURI       = flag.String("mysql_uri", "", "URL of a MySQL database to clone the log into. The DB should contain only one log.")
 	writeBatchSize = flag.Uint("write_batch_size", 1024, "The number of leaves to write in each DB transaction.")
 	workers        = flag.Uint("workers", 4, "The number of worker threads to run in parallel to fetch entries.")
 )
 
+const (
+	height = 8
```
AlCutter, line 43: nit: maybe `tileHeight` just to be explicit?

AlCutter (approved):

clone/cmd/sumdbclone/sumdbclone.go
mhutchinson: Done.

PR #456, opened 2021-08-23
ethwit - Securing install of pipenv and ganache

sinasabet81:
- for pipenv: the dockerfile (which is a reproducible install thanks to its hash) already includes pipenv, so we don't need to reinstall it using pip
- for ganache, we can add a yarn.lock, which secures the installed version of ganache as well as its deps and sub-deps


smeiklej (approved):
Looks good!

pavelkalinnikov (approved):

Issue #457, opened 2021-08-23
[Serverless] Split up witness PR handling


AlCutter:
Currently, the experimental witness PR action code tries to do everything (validate and merge, etc.) on the PR branch itself, but this gets tricky when the PR is coming from a fork.

Instead, we should just validate that the incoming witness PR is good, allow it to merge, and then update `checkpoint.witnessed` on master.

PR #458, opened 2021-08-23
[Serverless] Replace witness PR validator

AlCutter:
Trim down what the witness PR action is attempting to do so that we're only validating that the PR is good.

Part of #457


serverless/deploy/github/witness_pr/entrypoint.sh
```diff
@@ -11,26 +11,28 @@ set -e
...
 
     while IFS= read -r f; do
             f=$(readlink -f -n ${f})
-            if [[ ${f} = ${LOG_CP} ]]; then
-                echo "::error: file=${f} Modifications to checkpoint not allowed, please use checkpoint.witnessed"
-                has_witnessed_checkpoint=1
-            elif [[ ${f} = ${MODIFIED_CP} ]]; then
-                echo "::debug:Found checkpoint.witnessed"
+            if [[ ${f} = ${LOG_DIR}/checkpoint ]]; then
+                echo "::error: file=${f} Modifications to checkpoint not allowed, please add files to the witnessed/ directory."
```
mhutchinson, line 27: This says to add to the `witnessed` directory, but the check below is for `witness/*` ?

Whichever is the correct directory should be stated in the docs at `serverless/deploy/github/witness_pr/README.md`.

mhutchinson: OK, from looking at other PRs it seems that this doc is wrong but the code is right below. Approving this PR with the expectation this docs and the README will be clarified before merging.

mhutchinson (approved):

serverless/deploy/github/witness_pr/entrypoint.sh
AlCutter: good spot, fixed - I'm going to blame that on muscle memory :)

PR #459, opened 2021-08-23
[Serverless] Add a new combine-witness-signatures action

AlCutter:
Adds a new action intended to be used to combine witnessed checkpoints.
This should run on `merge to master` events, and will combine witnessed checkpoints from the `witness/` directory with either the one present in `checkpoint` or `checkpoint.witnessed`, writing the output to `checkpoint.witnessed`.

Part of #457


serverless/deploy/github/combine_witness_signatures/combine_witness_signatures.go
```diff
@@ -0,0 +1,209 @@
...
+	// if we have sufficient witness cosigs to promote checkpoint to
+	// checkpoint.witnessed
+	if cpwNote == nil || cpNote.Text != cpwNote.Text {
+		out, err = mergeCheckpoints(cpRaw, ans, *numRequired, logSigV, witSigV, false)
+		if err != nil {
+			glog.Warningf("Failed to merge with new checkpoint file: %v", err)
+		} else {
+			// Since we've just promoted the new checkpoint and merged any compatible
+			// cosignatures, we can delete all additional Checkpoints:
+			if err := os.RemoveAll(witnessDir); err != nil {
```
mhutchinson, line 92: Should this be guarded by `delete_consumed` too? That flag seems a little niche otherwise. With it also guarding this behaviour its behaviour is far simpler to understand.

serverless/deploy/github/combine_witness_signatures/combine_witness_signatures.go
```diff
@@ -0,0 +1,209 @@
...
+var (
+	storageDir     = flag.String("storage_dir", "", "Root directory of the log.")
+	logKey         = flag.String("log_public_key", "", "Log's public key.")
+	witnessKeys    = flag.String("witness_public_key_files", "", "One or more space-separated globs matching all the known witness keys.")
+	numRequired    = flag.Uint("required_witness_sigs", 0, "Specifies the minimum number of valid witness signatures required to be present on the output to consider the merge successful.")
+	output         = flag.String("output", "", "Output file to write combined checkpoint to.")
+	deleteConsumed = flag.Bool("delete_consumed", true, "Whether to delete files containing merged checkpoints.")
+)
+
+func main() {
```
mhutchinson, line 49: The behaviour for promoting/merging checkpoints in here would benefit from some tests. There are a sufficient number of moving parts that even with two sets of eyes on it I couldn't rule out the possibility of bugs.

mhutchinson (approved):

serverless/deploy/github/combine_witness_signatures/combine_witness_signatures.go
AlCutter: yeah, I agree - my plan was do the broad strokes to get it working, and then factor funcs out and add tests

serverless/deploy/github/combine_witness_signatures/combine_witness_signatures.go
AlCutter: This is a good question - the thinking here is that since we've just promoted (and combined any possible cosigs)`checkpoint` there can be nothing useful left in the `witness` directory, so we delete *everything* including non-consumed.

In the other branch, we only delete the consumed files because, conversely, there *may* be some number of useful cosigs over `checkpoint` there but which are insufficient to meet the threshold for promoting `checkpoint` to `checkpoint.witnessed`.

I've added a comment explaining the absence of deletion in the "cosign checkpoint.witnessed" branch.

PR #460, opened 2021-08-23
[Serverless] Fix up feeder

AlCutter:
- Update location to store witnessed checkpoints

Part of #457

### DIFF
--- a/serverless/cmd/feeder/github/main.go
+++ b/serverless/cmd/feeder/github/main.go
@@ -177,14 +177,15 @@ func feedOnce(ctx context.Context, opts *options, forkRepo *git.Repository, ghCl
 
 	glog.V(1).Infof("CP after feeding:\n%s", string(wRaw))
 
-	branchName := fmt.Sprintf("refs/heads/witness_%s", hex.EncodeToString(wCp.Hash))
+	hexHash := hex.EncodeToString(wCp.Hash)
+	branchName := fmt.Sprintf("refs/heads/witness_%s", hexHash)
 	deleteBranch, err := gitCreateLocalBranch(forkRepo, headRef, branchName)
 	if err != nil {
 		return fmt.Errorf("failed to create git branch for PR: %v", err)
 	}
 	defer deleteBranch()
 
-	outputPath := filepath.Join(opts.logPath, "checkpoint.witnessed")
+	outputPath := filepath.Join(opts.logPath, "witness", fmt.Sprintf("checkpoint_%s", hexHash))
 	msg := fmt.Sprintf("Witness checkpoint@%v", wCp.Size)
 	if err := gitCommitFile(workTree, outputPath, wRaw, msg, opts.gitUsername, opts.gitEmail); err != nil {
 		return fmt.Errorf("failed to commit updated checkpoint.witnessed file: %v", err)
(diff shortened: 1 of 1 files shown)


serverless/cmd/feeder/github/main.go
```diff
@@ -177,14 +177,15 @@ func feedOnce(ctx context.Context, opts *options, forkRepo *git.Repository, ghCl
...
+	hexHash := hex.EncodeToString(wCp.Hash)
+	branchName := fmt.Sprintf("refs/heads/witness_%s", hexHash)
 	deleteBranch, err := gitCreateLocalBranch(forkRepo, headRef, branchName)
 	if err != nil {
 		return fmt.Errorf("failed to create git branch for PR: %v", err)
 	}
 	defer deleteBranch()
 
-	outputPath := filepath.Join(opts.logPath, "checkpoint.witnessed")
+	outputPath := filepath.Join(opts.logPath, "witness", fmt.Sprintf("checkpoint_%s", hexHash))
```
mhutchinson, line 188: Won't this cause conflicts when merged into `master`? If I'm reading this right, all witnesses seeing the same checkpoint will create their own cosigned version called `witness/checkpoint_${LOG_ROOT_HASH}` on their own named branch. But `witness/checkpoint_${LOG_ROOT_HASH}` will be the same for all checkpoints, so when merged to master there will be a conflict. If numRequiredSigs > 1 then we'll never get the right number?

I'm guessing that I've missed some logic elsewhere that renames this file when merging to master, but I'll leave this comment and then find the other part of the machine that does that...

AlCutter: Yeah, with my noddy set up I don't hit that because the combine-on-master action tidies everything up fairly quickly, but in cases where more than 1 cosign was needed to promote you could certainly have collisions.

Just added a commit which tries to make the ID stable for a given CP+set-of-cosigs (I wanted a stable ID to avoid the feeder accidentally creating new PRs and branches all over the place)

serverless/cmd/feeder/github/main.go
```diff
@@ -201,6 +203,18 @@ func feedOnce(ctx context.Context, opts *options, forkRepo *git.Repository, ghCl
 	return createPR(ctx, opts, ghCli, fmt.Sprintf("Witness @ %d", wCp.Size), opts.feederRepo.owner+":"+branchName, "master")
 }
 
+// cpID returns a stable identifier for a given checkpoint and set of known signatures.
+// This is used as a branch name, and for generating a file name for the witness PR.
+func wcpID(n note.Note) string {
+	bits := []string{n.Text}
```
mhutchinson, line 209: Maybe parts or components instead of `bits`? I genuinely thought you were going to do some bitwise magic here to make this work.

serverless/cmd/feeder/github/main.go
```diff
@@ -201,6 +203,18 @@ func feedOnce(ctx context.Context, opts *options, forkRepo *git.Repository, ghCl
 	return createPR(ctx, opts, ghCli, fmt.Sprintf("Witness @ %d", wCp.Size), opts.feederRepo.owner+":"+branchName, "master")
 }
 
+// cpID returns a stable identifier for a given checkpoint and set of known signatures.
```
mhutchinson, line 206: Maybe nit, maybe important: this says a "set of", but sets are unordered and there is no sorting of the sigs here, hence the identifier generated doesn't look like it is stable if the sigs are in different orders.

Options::
 * s/set/list
 * sort the sigs
 * add docs here if `n.Sigs` is already sorted in some stable way

mhutchinson (approved):

serverless/cmd/feeder/github/main.go
AlCutter: Should be stably sorted, added a comment to that effect.

serverless/cmd/feeder/github/main.go
AlCutter: done

Issue #464, opened 2021-08-26
[Serverless] Break out log and distributor roles


AlCutter:
The serverless github action stuff supports both log and (nascent) witness distributor activity, these should be broken out so it's clear/easy to see how one might choose to setup and run one or both of these things both independently or even in the same repo if desired.

PR #468, opened 2021-08-31
[Serverless] Update feeder to support distributor

AlCutter:
Paired with #467 and part of #464


mhutchinson (approved):
Nice, this addresses a lot of my confusion from when this code was added about what the log repo was for. Calling it distributor makes it a lot clearer.

PR #470, opened 2021-09-01
[Formats] Rename Ecosystem checkpoint field to Origin

AlCutter:
The Ecosystem is a properly of the log which is a member of that ecosystem, whereas the checkpoint is the embodiment of a claim made by that log and as such should be explicit about its origin.

This change renders checkpoints signed by the "wrong" log key invalid, and mitigates a class of targeted attack when using witnesses to protect against split views.


AlCutter:
There are one or two more references to `Ecosystem` in the GitHub actions in this repo, but since they're actually separate modules I'll need to bump them separately once this PR has merged.

formats/log/README.md
```diff
@@ -45,25 +45,42 @@ signatures, the format itself is not restricted to this scheme.
...
-data to the checkpoint, however its format must conform to the sumdb signed
-note spec (e.g. it must not contain blank lines.)
+
+* `<Origin string>` should be a unique identifier for the log identity which issued the checkpoint.
+  The exact format of this is left undefined, but examples of types of value to put here
+  are the log's URL, a base64 encoded hash of its public key, etc.
+
+  The presence of this identifier forms part of the log claim, and prevents witness
+  signatures being moved between log checkpoints with identical contents, preventing a
+  particular type of targetted attack.
```
smeiklej, line 62: I think this is just spelled "targeted" but maybe that's an American thing?

formats/log/README.md
```diff
@@ -45,25 +45,42 @@ signatures, the format itself is not restricted to this scheme.
...
+  data to the checkpoint, however its format must conform to the sumdb signed
+  note spec (e.g. it must not contain blank lines.)
 
 > Note that golang sumdb implementation is already compatible with this
 `[otherdata]` extension (see
 <https://github.com/golang/mod/blob/d6ab96f2441f9631f81862375ef66782fc4a9c12/sumdb/tlog/note.go#L52>).
 
-The first signature on a checkpoint must be from the log which issued it.
+The first signature on a checkpoint should be from the log which issued it, but there MUST NOT
+be more than one signature from a log identity present on the checkpoint ()
```
smeiklej, line 80: Not sure what's happened with the empty brackets?

smeiklej (approved):
Two potential typos but otherwise looks good :).

formats/log/README.md
AlCutter: fixxed

formats/log/README.md
AlCutter: I think I was going to add a parenthetical about it being nonsensical to have a signature on a checkpoint from a log other than the one identified by the zeroth line, but that's probably not really necessary so I've ding'd 'em.

paulmattei (approved):

formats/log/README.md
```diff
@@ -45,25 +45,42 @@ signatures, the format itself is not restricted to this scheme.
...
-data to the checkpoint, however its format must conform to the sumdb signed
-note spec (e.g. it must not contain blank lines.)
+
+* `<Origin string>` should be a unique identifier for the log identity which issued the checkpoint.
+  The exact format of this is left undefined, but examples of types of value to put here
+  are the log's URL, a base64 encoded hash of its public key, etc.
+
+  The presence of this identifier forms part of the log claim, and prevents witness
+  signatures being moved between log checkpoints with identical contents, preventing a
+  particular type of targeted attack.
```
mhutchinson, line 62: This is true, but expanding this won't do us any harm. Something a little closer to a compelling explanation for me would be something like:

Including a unique log string in the checkpoint prevents distinct logs from creating checkpoints with the same contents. This prevents signatures from being transferred between checkpoints. As an example of an attack that this mitigates: a witness signature on a checkpoint confirms that the log's append-only property has been verified as of this checkpoint. If another log creates the same checkpoint, the witness signature can be used on that checkpoint which implies a verification check that may never have happened.

mhutchinson (approved):

formats/log/README.md
AlCutter: I'm not at all against sharing the details, but I'm not sure whether this is the right place to put them - if I'm implementing this, I suspect that at a first pass I just want to know what the fields are and how I should use them, later on I might be interested to know why.

Would having these sorts of details in an issue or appendix section be better? (Maybe I should also remove the "preventing a particular type..." bit too?)

mhutchinson: Yeah I think dropping that removes the slippery slope into explaining the whole thing here. To keep this short and still accurate, perhaps "The presence of this identifier forms part of the log claim, and means that no two logs can create byte-equivalent checkpoints.".

Witnesses, signatures, and all of that can then be moved elsewhere.

formats/log/README.md
```diff
@@ -45,25 +45,42 @@ signatures, the format itself is not restricted to this scheme.
...
 `[otherdata]` extension (see
 <https://github.com/golang/mod/blob/d6ab96f2441f9631f81862375ef66782fc4a9c12/sumdb/tlog/note.go#L52>).
 
-The first signature on a checkpoint must be from the log which issued it.
+The first signature on a checkpoint should be from the log which issued it, but there MUST NOT
+be more than one signature from a log identity present on the checkpoint.
 
 **Differences from sumdb root:**
-The sumbdb note has `go.sum database tree` as its ecosystem string.
+The sumbdb note has `go.sum database tree` as its `Origin`.
```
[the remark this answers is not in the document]
AlCutter: removed the section

formats/log/README.md
AlCutter: Reworded and remove the extra stuff for now - we can figure out where to put that later.

PR #475, opened 2021-09-03
Reordered parameters to ParseCheckpoint

mhutchinson:
This makes it much easier in the common case that there aren't additional verifiers


AlCutter (approved):
Yeah, this is better

PR #476, opened 2021-09-06
Perform checks on Origin consistently within FT demo

mhutchinson:
This change identified a few tests that were still passing using the old origin string that was replaced in #470. I think this is compelling that reusing the same parsing code reduces bugs.

### DIFF
--- a/formats/log/note.go
+++ b/formats/log/note.go
@@ -20,17 +20,8 @@ import (
 	"golang.org/x/mod/sumdb/note"
 )
 
-// CheckpointNote is a parsed Checkpoint Note. This embeds the Note object to
-// allow clients to find any signatures, and also provides access to the
-// parsed Checkpoint and unparsed OtherData.
-type CheckpointNote struct {
-	*note.Note
-	Checkpoint *Checkpoint
-	OtherData  []byte
-}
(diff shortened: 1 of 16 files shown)


formats/log/note.go
```diff
@@ -29,27 +36,30 @@ import (
 // checkpoint is returned. The underlying note is always returned where possible.
 // The signatures on the note will include the log signature if no error is returned,
 // plus any signatures from otherVerifiers that were found.
-func ParseCheckpoint(chkpt []byte, origin string, logVerifier note.Verifier, otherVerifiers ...note.Verifier) (*Checkpoint, *note.Note, error) {
+func ParseCheckpoint(chkpt []byte, origin string, logVerifier note.Verifier, otherVerifiers ...note.Verifier) (*Checkpoint, *CheckpointNote, error) {
```
AlCutter, line 39: Did you consider augmenting `CheckpointNote` struct to also contain the `*Checkpoint` and just return that?

mhutchinson: I've pushed a new commit that does this. I have mixed feelings. Let me know what you think.

AlCutter (approved):

PR #477, opened 2021-09-06
minor order of operations edit

paulmattei:
Example code prints public then private, which is ever so slightly unexpected given suggested bash is the other way around.


sigstore-feeder.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
+    "folders": [
+        {
+            "path": "."
+        }
+    ]
+}
```
mhutchinson, line 7: This file shouldn't be here.

mhutchinson (approved):

sigstore-feeder.code-workspace
paulmattei: ack - removed

mhutchinson (approved):

Issue #479, opened 2021-09-07
Use YAML for human-written config files


AlCutter:
JSON is a reasonable compromise for machine-generated-but-human-inspectable stuff, but it's not brilliant for human-authored configs. YAML is much better at this, and in theory at least, is a super-set of JSON so perhaps any existing JSON configs will continue to work.

mhutchinson:
The remaining json files aren't configs that we are in control of, or are for human configuration:

```
./witness/ethereum/package.json
./feeder/ethereum/package.json
./binary_transparency/firmware/testdata/firmware_metadata.json
./binary_transparency/firmware/testdata/firmware_statement.json
./.github/workflows/link_checker_config.json
```

Well done team, we got 'em all!

PR #484, opened 2021-09-09
Switch a couple more config files to use YAML

AlCutter:
Switches two more config files to yaml:
- serverless `combine_witness_signatures` action
- Go witness

Towards #479


smeiklej (approved):
I think I missed the rationale behind this change somewhere, but anyway what's here seems fine to me.

PR #497, opened 2021-09-21
Extracted common configs into separate package

mhutchinson:
In the process of attempting to extract a utility to push the latest checkpoint from a witness to a distributor, I found myself once again copying and pasting these definitions. Having them be common allows fragments describing these entities to be easily copied into config files.

The exception to this clean extraction is the Log.ID which is not really a property of the log. I've added a TODO to come back to this.

### DIFF
--- a/serverless/config/log.go
+++ b/serverless/config/log.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 )
 
+// Log describes a verifiable log in a config file.
 type Log struct {
 	// ID is the user-chosen string used to refer to the log.
 	// This may be different across witnesses, distributors, etc.
@@ -37,6 +38,7 @@ type Log struct {
 	URL string `yaml:"URL"`
 }
 
+// Validate checks that the log configuration is valid.
 func (l Log) Validate() error {
 	if l.ID == "" {
 		return errors.New("missing field: ID")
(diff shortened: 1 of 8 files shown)


serverless/config/log.go
```diff
@@ -0,0 +1,56 @@
...
+// witnesses, etc.
+package config
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+)
+
+type Log struct {
```
AlCutter, line 26: This probably should have a struct comment on it since it's exported?

serverless/config/witness.go
```diff
@@ -0,0 +1,41 @@
...
+
+package config
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+)
+
+type Witness struct {
```
AlCutter, line 23: This too probably should have a struct comment on it since it's exported?

AlCutter (approved):

AlCutter:
You might need to add this package (without using it), then raise a separate PR to update the action submodules to use it (since they'll require a `go get github.com/google/trillian-examples` to see it).

serverless/config/log.go
mhutchinson: Done.

serverless/config/witness.go
mhutchinson: Done.

mhutchinson:
I've split out a separate PR for using the config structs. It makes me a bit nervous that this build didn't fail.

PR #499, opened 2021-09-22
Tidied up the feeder

mhutchinson:
This cleans up after changing from many witnesses to one, and avoids the need to fetch the checkpoint again after Update succeeds.


AlCutter (approved):
Thanks for the tidy!

PR #503, opened 2021-09-23
Yaml example fix

paulmattei:


paulmattei:
Thanks - and I'll remember to add more context notes next time, especially if they are landing for review first thing :)

PR #510, opened 2021-09-27
Proposal for log identifier

mhutchinson:
This is determinstic given an origin string and public key, and is guaranteed to contain no characters that could abuse filesystem or URL path handling.

### DIFF
--- a/formats/log/identifier.go
+++ b/formats/log/identifier.go
@@ -1,3 +1,17 @@
+// Copyright 2021 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


formats/log/identifier.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+// ID returns the identifier to use for a log given the Origin
+// and the public key. This is the ID used to find checkpoints
+// for this log at distributors, and that will be used to feed
+// checkpoints to witnesses.
+func ID(origin string, key []byte) string {
```
AlCutter, line 12: It feels a bit unexpected given the comment that this doesn't use the origin string passed in?

mhutchinson: We can optimize to remove the origin if we don't need it. I've added this signature to take both of these things for now so that we can continue to discuss what the implementation should be. It's clear to me that it certainly shouldn't be taking anything *more* than these though.

AlCutter (approved):
LGTM, we can experiment a bit and see how it goes - I'll update my witnesses & distrib'r

PR #514, opened 2021-09-29
Updated sumdb docker setup after json->yaml change

mhutchinson:
Also arranged for the default ID to be used, and added a little documentation about witnessing other logs.


sumdbaudit/witness/cmd/feeder/main.go
```diff
@@ -66,8 +66,13 @@ func main() {
 		}
 	}
 
+	lid := *logID
```
AlCutter, line 69: super nit: probably should be `lID` (although it's harder to read :/)

AlCutter (approved):

sumdbaudit/witness/cmd/feeder/main.go
mhutchinson: The Go gods might be cross with me, but I'll stick with `lid` - it parses so much easier.

PR #515, opened 2021-09-29
[serverless] Don't map ports to host, just to other containers

AlCutter:
This allows the running of multiple instances of the feeder/witness/distribute combo on the same machine without needing to change ports.

### DIFF
--- a/serverless/deploy/docker/witness/docker-compose.yaml
+++ b/serverless/deploy/docker/witness/docker-compose.yaml
@@ -21,8 +21,8 @@ services:
       - "--config_file=/witness-config/${WITNESS_CONFIG_FILE:-witness.config}"
       - "--logtostderr"
     restart: always
-    ports:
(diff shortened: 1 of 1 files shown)


serverless/deploy/docker/witness/docker-compose.yaml
```diff
@@ -21,8 +21,8 @@ services:
       - "--config_file=/witness-config/${WITNESS_CONFIG_FILE:-witness.config}"
       - "--logtostderr"
     restart: always
-    ports:
-      - "8100:8100"
+    expose:
```
mhutchinson, line 24: Doesn't this have the downside that if you want to bring up a witness and it actually work as its own distributor then its API isn't exposed? I feel like that's a more common default use-case than running multiple containers. Let's chat about this tomorrow, there must be a reasonable compromise here.

PR #516, opened 2021-09-29
[serverless/distro] Use fixed branch name per {witness,log} pair

AlCutter:
This should help keep the number of branches pushed to the distributor forks down, since they'll get reused for each subsequent checkpoint.

### DIFF
--- a/serverless/cmd/distribute/github/main.go
+++ b/serverless/cmd/distribute/github/main.go
@@ -18,8 +18,6 @@ package main
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


serverless/cmd/distribute/github/main.go
```diff
@@ -173,6 +171,16 @@ func distributeOnce(ctx context.Context, opts *options, repo github.Repository)
 	return repo.CreatePR(ctx, fmt.Sprintf("Witness %s@%d", opts.witSigV.Name(), wCp.Size), branchName)
 }
 
+func safeBranchChars(i rune) rune {
```
mhutchinson, line 174: I really want this to be called blanketyBlank.

mhutchinson (approved):

PR #518, opened 2021-10-04
Fixed typos in README

mhutchinson:
The most significant is that the config file for auto-merge didn't work.

### DIFF
--- a/serverless/deploy/github/distributor/README.md
+++ b/serverless/deploy/github/distributor/README.md
@@ -165,7 +165,7 @@ your github repo's page).
 ## Going further: automated PR merges
 
 Automating the merging incoming witness PRs can help to take some friction
-out of the process and help to get cosigned checpoints out to clients quicker
+out of the process and help to get cosigned checkpoints out to clients quicker
 than with manual review.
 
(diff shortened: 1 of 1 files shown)


serverless/deploy/github/distributor/README.md
```diff
@@ -224,7 +224,7 @@ jobs:
       - uses: actions-ecosystem/action-add-labels@v1
         with:
           labels: Automerge
-          number: ${{ steps.pr_metadata.outputs.pr }}
+          number: ${{ steps.find_pr_number.outputs.pr }}
```
AlCutter, line 227: Thanks for catching this - rather than change this line, though, could you rename the ID on L219 to be `pr_metadata` (this breakage was due to the manual merge I did between this file and the live config on `serverless-test`).

That edit should also fix this, but leaves the door open for other types of metadata to come through this channel too.

AlCutter (approved):

serverless/deploy/github/distributor/README.md
mhutchinson: Sure, done.

AlCutter (approved):

PR #519, opened 2021-10-04
[witness] no more optimistic signing

smeiklej:
Sign checkpoints only at the point at which we're sure they're consistent.

Oh, and I fixed a typo I noticed the other day :).

### DIFF
--- a/witness/golang/cmd/witness/internal/witness/witness.go
+++ b/witness/golang/cmd/witness/internal/witness/witness.go
@@ -145,11 +145,6 @@ func (w *Witness) Update(ctx context.Context, logID string, nextRaw []byte, proo
 	if err != nil {
 		return nil, fmt.Errorf("couldn't parse input checkpoint: %v", err)
 	}
-	// Optimistically sign the new checkpoint now.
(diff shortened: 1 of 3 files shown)


witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -206,6 +206,10 @@ func (w *Witness) Update(ctx context.Context, logID string, nextRaw []byte, proo
 		}
 		// If the proof is good store nextRaw and the new range.
 		r := []byte(log.Proof(nextRange).Marshal())
+		signed, err := w.signChkpt(nextNote)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't sign input checkpoint: %v", err)
+		}
 		if err := w.setChkptData(tx, logID, signed, r); err != nil {
 			return nil, fmt.Errorf("failed to store new checkpoint: %v", err)
 		}
```
AlCutter, line 215: Could potentially factor this out into a separate func called from the 3 locations?

AlCutter (approved):

PR #537, opened 2021-11-02
[Feeder] Factor out some common code

AlCutter:
The intention is to try to share some of the error handling for corner cases across feeders.

Probably worth reviewing commit by commit.

### DIFF
--- a/serverless/cmd/feeder/main.go
+++ b/serverless/cmd/feeder/main.go
@@ -30,9 +30,10 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/google/trillian-examples/formats/log"
+	"github.com/google/trillian-examples/internal/feeder"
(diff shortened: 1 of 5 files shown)


serverless/cmd/feeder/main.go
```diff
@@ -90,42 +91,41 @@ func feedLog(l config.Log, w wit_http.Witness, timeout time.Duration, interval t
 		return fmt.Errorf("invalid LogURL %q: %v", l.URL, err)
 	}
 	f := newFetcher(lURL)
+	// TODO(al): make this configurable
```
mhutchinson, line 94: I think this TODO can be dropped. There doesn't seem to be any pretense of flexibility of hasher in the serverless implementation at the moment.

mhutchinson (approved):

PR #548, opened 2021-11-09
Improve witness status code responses

AlCutter:
- tweaked the returned `status.Code` for some conditions
- added a bit of logging/passing HTTP status text through into `error`s
- fixed bug where CP wasn't being written as comments suggested

### DIFF
--- a/witness/golang/cmd/witness/internal/witness/witness.go
+++ b/witness/golang/cmd/witness/internal/witness/witness.go
@@ -133,7 +133,7 @@ func (w *Witness) GetCheckpoint(logID string) ([]byte, error) {
 // version of nextRaw if the update was applied.
 //
 // If an error occurs, this method will generally return an error with a status code:
-// - codes.NodeFound if the log is unknown
+// - codes.NotFound if the log is unknown
 // - codes.InvalidArgument for general bad requests
 // - codes.AlreadyExists if the checkpoint is smaller than the one the witness knows
 // - codes.FailedPrecondition if the checkpoint is inconsistent with the one the witness knows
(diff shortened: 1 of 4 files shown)


witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -126,8 +127,16 @@ func (w *Witness) GetCheckpoint(logID string) ([]byte, error) {
...
 // Update updates the latest checkpoint if nextRaw is consistent with the current
-// latest one for this log. It returns the latest cosigned checkpoint held by
-// the witness, which is a signed version of nextRaw if the update was applied.
+// latest one for this log.
+//
+// It returns the latest cosigned checkpoint held by the witness, which is a signed
+// version of nextRaw if the update was applied.
+//
+// If an error occurs, this method will generally return an error with a status code:
+// - codes.NodeFound if the log is unknown
```
mhutchinson, line 136:
```suggestion
// - codes.NotFound if the log is unknown
```

mhutchinson (approved):
Did you review clients to see if they need updating for the new error codes?

witness/golang/cmd/witness/internal/witness/witness.go
AlCutter: _*cough*_ Signs of one too many Merkle implementations... :)

AlCutter:
> Did you review clients to see if they need updating for the new error codes?

Yup; nothing currently takes action based on the precise error type (and if that were to happen, it should probably be in the shared `/internal/feeder` package for consistency across feeder impls).

PR #552, opened 2021-11-18
Fix up the ECDSA verifier string format

AlCutter:
- Make the verifier string format be like standard note ones
- Rename from SigStoreECDSA -> ECDSA since this is now agreed to be the definition of note's algoID 2

### DIFF
--- a/internal/note/note_verifier.go
+++ b/internal/note/note_verifier.go
@@ -80,9 +80,9 @@ func (v *verifier) Verify(msg, sig []byte) bool {
 //   <key_name>+<key_hash>+<key_bytes>
 // Where
 //   <key_name> is a human readable identifier for the key, containing no whitespace or "+" symbols
-//   <key_hash> is a 32bit hash of the key bytes
 //   <key_bytes> is base64 encoded blob starting with a 0x02 (algECDSAWithSHA256) byte and followed
 //       by the DER encoded public key in SPKI format.
+//   <key_hash> is a 32bit hash of the key DER
 // e.g.:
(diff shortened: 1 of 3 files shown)


internal/note/note_verifier.go
```diff
@@ -71,25 +73,44 @@ func (v *verifier) Verify(msg, sig []byte) bool {
...
-// NewSigstoreECDSAVerifier creates a new note verifier for checking ECDSA signatures over SHA256 digests.
+// NewECDSAVerifier creates a new note verifier for checking ECDSA signatures over SHA256 digests.
 // This implementation is compatible with the signature scheme used by the Sigstore Rékor Log.
 //
 // The key is expected to be provided as a string in the following form:
-//   <key name> " " <Base64 encoded SubjectPublicKeyInfo DER>
+//   <key_name>+<key_hash>+<key_bytes>
+// Where
+//   <key_name> is a human readable identifier for the key, containing no whitespace or "+" symbols
+//   <key_hash> is a 32bit hash of the key bytes
```
mhutchinson, line 83: 32bit hash: is this more precisely the first 32 bits of a SHA256 over key_bytes?

mhutchinson (approved):

internal/note/note_verifier.go
AlCutter: Yep, that is technically what the implementation is, but I was being intentionally vague here to line up with the proposed change to the comments in note: https://go-review.googlesource.com/c/mod/+/364856 (e.g. folks shouldn't _rely_ on this being a truncated SHA256 over the DER).

I've tweaked the comment slightly though to reflect that it's the DER and not the `key_bytes`.

PR #555, opened 2021-11-19
Fix proof handling in Pixel BT feeder

AlCutter:


mhutchinson (approved):
Ah, that old rake in the grass.

PR #565, opened 2021-11-29
[Witness] Remove pubkey from feeders

AlCutter:
Having the public key around in feeders is a remnant from when the feeder/verifier/distributor roles were bundled together, but since the feeder's only job is to send checkpoints _into_ the witness it's unnecessary.

Removing it simplifies code & config.


mhutchinson (approved):
Woohoo

PR #569, opened 2021-12-03
[Serverless] Always write out checkpoint on update

AlCutter:
If the `output_checkpoint` flag is set, it's helpful to always create the file with the latest known checkpoint, regardless of whether it's just been updated.


smeiklej (approved):
LGTM!

PR #570, opened 2021-12-03
[Witness] Extract feeder implementations from their main files

AlCutter:
This allows alternative main files to be created for these feeders, e.g. TamaGo unikernels.

The delta appears large, but is almost entirely mechanical moving of existing code.

### DIFF
--- a/internal/feeder/pixelbt/convert.go
+++ b/internal/feeder/pixelbt/convert.go
@@ -25,7 +25,7 @@ import (
 	"unicode/utf8"
 )
 
-// convertToNote converts a the PixelBT checkpoint to a valid Note.
+// convertToNote converts a PixelBT checkpoint to a valid Note.
 //
 // Hopefully we won't need this for too long, and PixelBT will update their checkpoint format to make it
 // fully compatible.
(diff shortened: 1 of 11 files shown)


internal/feeder/pixelbt/convert.go
```diff
@@ -26,11 +25,11 @@ import (
 	"unicode/utf8"
 )
 
-// ConvertToNote converts a the PixelBT checkpoint to a valid Note.
+// convertToNote converts a the PixelBT checkpoint to a valid Note.
```
mhutchinson, line 28: `a` or `the`. Not introduced here, but I spotted it in this change.

mhutchinson (approved):
Update PR description from `witness implemenations` to `feeder implementations` for clarity. Otherwise LGTM.

internal/feeder/pixelbt/convert.go
AlCutter: a the done.

AlCutter:
> Update PR description from `witness implemenations` to `feeder implementations` for clarity. Otherwise LGTM.

done, ta

PR #582, opened 2022-01-14
Move ErrDupeLeaf to pkg directory and rename its package name.

jiggoha:
This error is currently in the `internal` dir, which means it cannot be used in serverless/experimental/gcp-log/.


serverless/cmd/sequence/main.go
```diff
@@ -26,12 +26,13 @@ import (
...
 
 	"github.com/google/trillian-examples/formats/log"
-	"github.com/google/trillian-examples/serverless/internal/storage"
 	"github.com/google/trillian-examples/serverless/internal/storage/fs"
 	"golang.org/x/mod/sumdb/note"
 
 	"github.com/golang/glog"
 	"github.com/transparency-dev/merkle/rfc6962"
+
+	stErrors "github.com/google/trillian-examples/serverless/pkg/storage/errors"
```
AlCutter, line 35: Did you consider moving it to `...pkg/log`?
I wonder if returning this concrete error should be part of the documented `Storage` interface there?

jiggoha: Good point. Done.

AlCutter (approved):

PR #585, opened 2022-01-27
Add Sequence function.

jiggoha:


serverless/experimental/gcp-log/function.go
```diff
@@ -12,27 +12,149 @@
...
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to parse Checkpoint: %q", err), http.StatusBadRequest)
+		return
+	}
+	client.SetNextSeq(cp.Size)
+
+	// sequence entries
+
+	h := rfc6962.DefaultHasher
+	it := client.GetObjects(ctx, d.EntriesDir)
```
AlCutter, line 109: Does anything delete pending entries once they're successfully sequenced, or is that for a later PR?

serverless/experimental/gcp-log/internal/storage/storage.go
```diff
@@ -51,8 +52,9 @@ type Client struct {
 	nextSeq uint64
 }
 
-// NewClient returns a Client which allows interaction with the log implemented on GCS.
-func NewClient(ctx context.Context, projectID string) (*Client, error) {
+// NewClient returns a Client which allows interaction with the log implemented
```
AlCutter, line 55: /implemented/stored/`?

serverless/experimental/gcp-log/function.go
```diff
@@ -12,27 +12,149 @@
...
+	// TODO(jayhou): validate that EntriesDir is only touching the log path.
+
+	var d struct {
+		Bucket string `json:"bucket"`
+		EntriesDir string   `json:"entriesDir"`
+		Origin     string `json:"origin"`
+	}
+
+	if err := json.NewDecoder(r.Body).Decode(&d); err != nil {
+		fmt.Printf("json.NewDecoder: %v", err)
```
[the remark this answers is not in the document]
jiggoha: I noticed I was using both, and wanted to make it consistent for now. `log` writes to stderr and `fmt` writes to stdout. In a later PR, I would like to add severity labels: https://cloud.google.com/functions/docs/monitoring/logging#writing_runtime_logs.

serverless/experimental/gcp-log/function.go
```diff
@@ -12,27 +12,149 @@
...
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
```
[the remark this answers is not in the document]
jiggoha: Done.

serverless/experimental/gcp-log/function.go
jiggoha: Are you referring to the directory `<rootDir>/leaves/pending/...`? I know in the Github Actions implementation, we created this directory [*] in order to write a temporary file which we remove after the full sequencing. I think we discussed that GCS does not have the concept of linking, so for the GCP implementation, I directly write the file without writing a temporary file first.

[*] https://github.com/google/trillian-examples/blob/ca80f135b82b9173e1bfbf45feec841edf01b5bf/serverless/internal/storage/fs/fs.go#L133

AlCutter: Ah, I was thinking of this: 

https://github.com/google/trillian-examples/blob/ca80f135b82b9173e1bfbf45feec841edf01b5bf/serverless/deploy/github/log/sequence_and_integrate/entrypoint.sh#L37

We should probably pop a pin in to think about this later - it might be better for the `sequence` tool to (optionally?) remove source leaf files one they've been sequenced since it knows for sure which ones have been.

wdyt?

AlCutter (approved):

PR #587, opened 2022-01-28
Integrate

jiggoha:
Fix bugs from testing GCF integrate function with `--initialise=false`.

Transform gcs.ErrObjectNotExist to os.ErrNotExist and add missing sequence return case.


serverless/pkg/log/integrate.go
```diff
@@ -176,15 +176,16 @@ func (tc tileCache) Visit(id compact.NodeID, hash []byte) {
...
 	tileKey := tileKey{level: tileLevel, index: tileIndex}
 	tile := tc.m[tileKey]
-	var err error
 	if tile == nil {
 		// We haven't see this tile before, so try to fetch it from disk
 		created := false
+
+		var err error
 		tile, err = tc.getTile(tileLevel, tileIndex)
+		if !os.IsNotExist(err) {
```
AlCutter, line 185: will this panic if `err` is nil?

jiggoha: Ah, good catch. Reverted this change.

AlCutter (approved):

PR #593, opened 2022-02-11
[Witness] Fix DB related issues

AlCutter:
Fixes a couple of issues with the witness' storage implementation:
1. The call to `Exec` to update checkpoints can silently fail, ensure we catch those errors
2. Errors caught by the fix to (1) were seen for concurrent update RPCs (where updates are for distinct logs), resulting in a `database locked` error. Avoid this situation entirely by limiting the number of concurrent connections.

Also simplify the SQL used for the update.

### DIFF
--- a/witness/golang/cmd/witness/internal/witness/witness.go
+++ b/witness/golang/cmd/witness/internal/witness/witness.go
@@ -155,6 +155,7 @@ func (w *Witness) Update(ctx context.Context, logID string, nextRaw []byte, proo
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create db tx: %v", err)
 	}
+	// Defer a rollback to clean up the TX if something fails.
 	defer tx.Rollback()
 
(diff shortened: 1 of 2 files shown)


witness/golang/cmd/witness/internal/witness/witness.go
```diff
@@ -155,6 +155,8 @@ func (w *Witness) Update(ctx context.Context, logID string, nextRaw []byte, proo
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create db tx: %v", err)
 	}
+	defer tx.Rollback()
```
mhutchinson, line 158:
```suggestion
        // Defer a rollback in case anything fails.
	defer tx.Rollback()
```

mhutchinson (approved):

witness/golang/cmd/witness/internal/witness/witness.go
AlCutter: done

PR #595, opened 2022-02-18
Updated docs on feeders

mhutchinson:
This expands the list to all known feeders, but suggests not running them individually as we have a better solution in the omniwitness.

### DIFF
--- a/witness/README.md
+++ b/witness/README.md
@@ -77,8 +77,8 @@ defined above.  Currently it contains implementations in the following languages
 ## Feeders
 
 Specific feeders are listed below, however potential witness operators are advised to
-simply deploy the [Omniwitness](golang/omniwitness) unless there is a compelling reason
-to limit the witnessed logs.
+simply deploy the [Omniwitness](golang/omniwitness) which contains all of these
+feeders, unless there is a compelling reason to limit the witnessed logs.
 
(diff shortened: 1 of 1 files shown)


witness/README.md
```diff
@@ -76,7 +76,11 @@ defined above.  Currently it contains implementations in the following languages
 
 ## Feeders
 
-* [Go SumDB](https://github.com/google/trillian-examples/tree/master/sumdbaudit/witness) :
-    a custom feeder for sum.golang.org that can be easily deployed with the generic golang witness implementation
-* [Serverless](https://github.com/google/trillian-examples/tree/master/serverless/cmd/feeder) :
-    a feeder command that acquires checkpoints from [serverless logs](https://github.com/google/trillian-examples/tree/master/serverless).
+Specific feeders are listed below, however potential witness operators are advised to
+simply deploy the [Omniwitness](golang/omniwitness) unless there is a compelling reason
```
AlCutter, line 80: Is it worth adding a parenthetical to explain that the OmniW is bundled with feeders and configs for all known logs?

AlCutter (approved):

witness/README.md
mhutchinson: I added a bit more here.

AlCutter (approved):

PR #598, opened 2022-02-28
Add Cloud Build step to write verifiable log entry to GCS.

jiggoha:


jiggoha:
Hi Al,

This is a first draft of what the changes to the cloudbuild config looks like. I have added two steps for logging just the witness image to demonstrate; please let me know if I understood your specs correctly. Particularly with the logging the architecture + OS, is this what you had in mind? (I don't actually quite understand why we need it.)

cloudbuild_docker.yaml
```diff
@@ -42,6 +43,52 @@ steps:
...
+      - '-E'
+      - '-c'
+      - >
+        # Special /workspace path:
+        # https://cloud.google.com/build/docs/configuring-builds/pass-data-between-steps#passing_data_using_workspaces
+        gcloud functions call create_gcs_object --data="{
+          \"bucket\":\"${_BUCKET}\",
+          \"entryContent\":\"${REPO_NAME},${COMMIT_SHA},$(cat /workspace/image-digest.txt),linux/amd64\",
+          \"entryPath\":\"${REPO_NAME}-linux-amd64\"
+        }" &&
```
jiggoha, line 79: FYI the quote escaping is a bit annoying here, but I couldn't find a way to avoid it. The outside double quotes allow the `$(cat ...)` command to be executed, and the inside/escaped double quotes are required by JSON.

AlCutter: Yeah, it's ok I've been down this bumpy road too :)

cloudbuild_docker.yaml
```diff
@@ -42,6 +43,52 @@ steps:
...
+  # the logs.
+  - name: 'gcr.io/cloud-builders/docker'
+    entrypoint: bash
+    args:
+      # Set ERR trap to be inherited by shell functions. Allows our bash command logs to show up in
+      # Cloud Console.
+      - '-E'
+      - '-c'
+      - >
+        docker pull gcr.io/$PROJECT_ID/witness &&
```
AlCutter, line 57: There might be a small issue here with what's being logged (related to your comment on the PR as a whole about OS/Platform etc.), but we can fix this later once you're happy with the log tooling etc.

Above, where the docker images are being built, they're using `buildx` to create images for all the platforms listed in the `_DOCKER_BUILDX_PLATFORMS` list (defined at the bottom of the file) - concretely, we'll currently build an image for `64bit x86` as well as `32bit armv7`.

I _think_ that when this `docker pull` runs, it'll implicitly pull the image for the platform _its_ running on which is likely to be `amd64` today (but could potentially one day be an arm based machine), and uses that to create the `image-digest.txt` file.

But then we use the contents from that `image-digest.txt` file for both `x86` and `armv7` entries in the log, so one of those entries is likely to be incorrect (I'm assuming that the claim these leaves represent is `"The TrustFabric team claims that the docker image from ${container registry} with hash ${hash} was built from ${dockerfile} in ${repo} at commit hash ${hash} on ${platform}"` - when you're happy with the GCS log it might be worth thinking this assumption through! :) )

Anyway, I suspect you might need to loop over the `_DOCKER_BUILDX_PLATFORMS` and re-pull the docker image each time using the `--platform` flag, maybe something like:

```bash
for platform in $(echo $_{DOCKER_BUILDX_PLATFORMS} | tr , " "); do
  docker pull --platform=${platform} gcr.io/${PROJECT_ID}/witness
  docker images --format '{{.Digest}}' gcr.io/${PROJECT_ID}/witness) > /workspace/image-digest-${platform}.txt
  gcloud functions call ...
done
```

but also just not worrying about it too much until we've looked at the claim and arrived at what we think _should_ be in there is fine too!

jiggoha: As we discussed offline, we will not focus on the contents of the log yet, thus I have removed the platform-related code. My plan is to get this PR in as an MVP for the Cloud Build trigger, deploy serverless log GCF functions, then add their invocation here in another PR.

AlCutter (approved):

PR #599, opened 2022-03-03
Improve feeder interfaces and timeout handling

mhutchinson:
All feeders now take a http client used for all http connections. Furthermore, they all take the interface definition for a witness instead of the http client implementation. This allows different implementations to be passed in

All clients configure the http client with a timeout. Previously, it doesn't look like the timeout was actually being used in many cases.

### DIFF
--- a/internal/feeder/serverless/serverless_feeder.go
+++ b/internal/feeder/serverless/serverless_feeder.go
@@ -30,12 +30,11 @@ import (
 	"github.com/transparency-dev/merkle/rfc6962"
 
 	i_note "github.com/google/trillian-examples/internal/note"
-	wit_http "github.com/google/trillian-examples/witness/golang/client/http"
 )
 
 // FeedLog periodically feeds checkpoints from the log to the witness.
 // This function returns once the provided context is done.
-func FeedLog(ctx context.Context, l config.Log, w wit_http.Witness, timeout time.Duration, interval time.Duration) error {
+func FeedLog(ctx context.Context, l config.Log, w feeder.Witness, c *http.Client, interval time.Duration) error {
 	lURL, err := url.Parse(l.URL)
 	if err != nil {
 		return fmt.Errorf("invalid LogURL %q: %v", l.URL, err)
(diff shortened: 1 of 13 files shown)


internal/feeder/serverless/serverless_feeder.go
```diff
@@ -84,9 +83,30 @@ func FeedLog(ctx context.Context, l config.Log, w wit_http.Witness, timeout time
 // TODO(al): factor this stuff out and share between tools:
 
 // newFetcher creates a Fetcher for the log at the given root location.
-func newFetcher(root *url.URL) client.Fetcher {
-	get := getByScheme[root.Scheme]
-	if get == nil {
+// If the scheme is http/https then the client provided will be used.
+func newFetcher(c *http.Client, root *url.URL) client.Fetcher {
```
AlCutter, line 87: Would it be better to just have `FeedLog` take a `client.Fetcher` directly, and move this out to be the responsibility of the caller?

Passing a "maybe" `Client` in here is showing that we should probably get around to that `TODO` above, and figure out what this should look like as we factor it out.

mhutchinson: We _could_ do that. The reason I didn't is that all of the FeedLog implementations have the same interface as of this PR. Previously they all did except for the sumdb one, which was missing the timeout. I could change this one, but then it would be the odd one out.

I suggest we leave this as-is in this PR. I think what we have here is strictly better than what we had before in terms of functionality and consistency (passing a timeout into methods where it was then ignored seems really suspect). Future work should address this TODO and probably do so consistently across all of the feeders. Making that `Fetcher` interface live somewhere common and using it instead of the `*http.Client` on all feeders is a likely outcome.

AlCutter: > Making that `Fetcher` interface live somewhere common and using it instead of the `*http.Client` on all feeders is a likely outcome.

Yeah, this sounds like the sort of thing.
Worth expanding the TODO slightly as a memory jogger?

AlCutter (approved):

PR #604, opened 2022-03-16
serverless/client: Don't panic when tile is unreachable

Foxboron:
Happens if we specify a non-existing "to-size" in a consistency proof

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/serverless/client/client.go
+++ b/serverless/client/client.go
@@ -254,12 +254,12 @@ func (n *nodeCache) GetNode(ctx context.Context, id compact.NodeID) ([]byte, err
 		n.tiles[tKey] = *tile
 	}
 	nodeKey := int(api.TileNodeKey(nodeLevel, nodeIndex))
-	if nodeKey > len(t.Nodes) {
-		return nil, fmt.Errorf("node %v (tile coords [%d,%d]/[%d,%d]) unreachable", id, tileLevel, tileIndex, nodeLevel, nodeIndex)
+	if l := len(t.Nodes); nodeKey >= l {
+		return nil, fmt.Errorf("node %v (tile coords [%d,%d]/[%d,%d], key %d) outside populated tile nodes (%d)", id, tileLevel, tileIndex, nodeLevel, nodeIndex, nodeKey, l)
 	}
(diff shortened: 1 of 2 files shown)


AlCutter:
Thanks for the fix @Foxboron!

Incidentally, I'm curious how this happened - ideally the clients wanting the consistency proof should have signed checkpoints committing to both the smaller and larger sizes, so _in theory at least_ the nodes must have been written out. 

My guess would be something like the frontends serving the log state files are distributed and the state hadn't replicated to all frontends, so your client saw a newer/larger checkpoint from frontend A, but asked frontend B for tiles but B had not yet caught up?

Foxboron:
A better check would be to just validate the input data for the known tree size instead of just claiming it's unreachable.

AlCutter:
> A better check would be to just validate the input data for the known tree size instead of just claiming it's unreachable.

We could enforce that the `client` binary wants to verify the two `checkpoint` files before trying to build the proof between them, that would get rid of user error causing the issue but at a slight increase in complexity of having to keep those `checkpoint` artefacts around.  However the server-skew scenario can still easily happen so we'll always have the potential of a 404 looking for tile files...

serverless/client/client.go
```diff
@@ -253,7 +253,11 @@ func (n *nodeCache) GetNode(ctx context.Context, id compact.NodeID) ([]byte, err
 		t = *tile
 		n.tiles[tKey] = *tile
 	}
-	node := t.Nodes[api.TileNodeKey(nodeLevel, nodeIndex)]
+	nodeKey := int(api.TileNodeKey(nodeLevel, nodeIndex))
+	if nodeKey > len(t.Nodes) {
+		return nil, fmt.Errorf("node %v (tile coords [%d,%d]/[%d,%d]) unreachable", id, tileLevel, tileIndex, nodeLevel, nodeIndex)
```
AlCutter, line 258: nit: `unreachable` -> `not found`?

AlCutter (approved):

Foxboron:
Sorry for being slow, thanks for the fixup :)

PR #605, opened 2022-03-17
Update USB Armory related deps to new location

AlCutter:


AlCutter:
I think this should be updated with the new references too now, PTAL?

PR #606, opened 2022-03-21
sumdbconfig

mhutchinson:
Make SumDB Feeder configured similarly to the others

This feeder was implemented before the others, and at the time it seemed like a good idea to have it take the URL from the key name. This now takes the URL explicitly, which makes it consistent with the others. The config for this feeder in the omniwitness now lives in a config file.


AlCutter (approved):
Final commit LGTM for rebase!

PR #610, opened 2022-03-23
Run 'yarn upgrade' to fix some security warnings

mhutchinson:
We're getting an increasing number of security warnings. Hopefully this will address some of them. I haven't tested this, hopefully the presubmits will run some tests to ensure this hasn't changed functional behaviour.


smeiklej (approved):
LGTM!  Curious to see how many this does address.

Issue #614, opened 2022-03-25
[Witness/Armory] Add some form of durable storage


AlCutter:

mhutchinson:
@AlCutter - anything else you're planning on doing for this? Seems done to me!

AlCutter:
Yeah, I guess we can close it - I left it open just in case we wanted to wire it in for storing user-tweakable configs etc., but that could easily be its own issue too.

PR #615, opened 2022-03-25
[Witness/Armory] Add support for reading/writing blocks to SD/MMC

AlCutter:
Towards #614

### DIFF
--- a/witness/golang/omniwitness/usbarmory/main.go
+++ b/witness/golang/omniwitness/usbarmory/main.go
@@ -31,7 +31,9 @@ import (
 	"golang.org/x/mod/sumdb/note"
 	"golang.org/x/sync/errgroup"
 
-	_ "github.com/usbarmory/tamago/board/f-secure/usbarmory/mark-two"
+	"github.com/google/trillian-examples/witness/golang/omniwitness/usbarmory/storage"
+
+	usbarmory "github.com/usbarmory/tamago/board/f-secure/usbarmory/mark-two"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


witness/golang/omniwitness/usbarmory/main.go
```diff
@@ -45,10 +47,32 @@ const (
...
+	flag.Set("stderrthreshold", "INFO")
 	flag.Parse()
+
+	if err := usbarmory.MMC.Detect(); err != nil {
+		glog.Exitf("Failed to detect MMC: %v", err)
+	}
+
+	_ = storage.Device{
+		Card: usbarmory.MMC,
+	}
```
mhutchinson, line 74: If this has an effect, then a comment would be useful. If not, a TODO or deleting it would be cleaner (in case you end up side-tracked somewhere before follow up PRs are in)

witness/golang/omniwitness/usbarmory/storage/internal.go
```diff
@@ -0,0 +1,88 @@
...
+// by the USB Armory.
+package storage
+
+import "github.com/usbarmory/tamago/soc/imx6/usdhc"
+
+var (
+	// MaxTransfer is the largest transfer we'll attempt.
+	// If we're asked to read or write more data than can fit into available DMA memeory
+	// we'll had a bad time, so we'll chunk into requests of at most MaxTransfer bytes.
+	MaxTransfer = 32 * 1024
```
mhutchinson, line 28: Consider `MaxTransferBytes` to be clear about the unit.

witness/golang/omniwitness/usbarmory/storage/internal.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// WriteBlocks writes the data in b to the device blocks starting at the given block address.
+// If the final block to be written is partial, it will be padded with zeroes to ensure that
+// full blocks are written.
+func (d *Device) WriteBlocks(lba uint, b []byte) error {
+	if len(b) == 0 {
+		return nil
+	}
+	ci := d.Card.Info()
```
mhutchinson, line 49: In both this and the Read method, I think it would be clearer to do `blockSize := d.BlockSize()` and use that. It doesn't look like `ci` is used for anything other than querying that.

witness/golang/omniwitness/usbarmory/storage/internal.go
```diff
@@ -0,0 +1,88 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build usbarmory
+// +build usbarmory
+
+// Package storage provides support for using the SD/eMMC storage provided
+// by the USB Armory.
```
mhutchinson, line 19: Perhaps some comment about this being extremely raw access, and it being up to the code to make sure it doesn't delete itself if it is installed on the device?

mhutchinson (approved):

witness/golang/omniwitness/usbarmory/storage/internal.go
AlCutter: done

witness/golang/omniwitness/usbarmory/storage/internal.go
AlCutter: Good call, done

witness/golang/omniwitness/usbarmory/storage/internal.go
AlCutter: done

PR #616, opened 2022-03-25
[Witness/Armory] Add a journal implementation

AlCutter:
Towards #614


witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+const entryHeaderSize = 4 + 4 + 8 + 32
+
+// Size returns the number of bytes used by this entry record.
+func (e *entry) Size() int {
+	return entryHeaderSize + len(e.Data)
+}
+
+// OpenJournal returns a new journal structure for interacting with a journal stored in the
+// [start, start+length) range of blocks accessible via dev.
+func OpenJournal(dev BlockReaderWriter, start, length uint) (*Journal, error) {
```
mhutchinson, line 83: A wee discussion about the expectations around locking / ownership responsibilities would be very useful. I imagined that this would have a mutex in it, and the contract would be that there should be only one journal covering any particular range. However, given the lack of a mutex in here, either a) it's not thread-safe, or b) all locking and safety is performed at a higher level.

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+	j := &Journal{
+		dev:           dev,
+		start:         start,
+		length:        length,
+		maxWriteBytes: (length*dev.BlockSize())/3 - entryHeaderSize,
```
mhutchinson, line 88: This 3 should be a constant (`minEntryCount` ?). It looks well magic here. It's also quite carefully chosen, so explaining that would be great for future visitors. In particular, that there can be _more_ entries than this is non-obvious.

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	if err := j.init(); err != nil {
+		return nil, err
+	}
+
+	return j, nil
```
mhutchinson, line 95:
```suggestion
	return j, j.init()
```

Unless you're worried about clients using `j` even if there's an error.

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// Journal implements a record-based format which provides a resilient storage.
+type Journal struct {
+	dev           BlockReaderWriter
+	start         uint
+	length        uint
+	current       entry
+	nextBlock     uint
+	maxWriteBytes uint
```
mhutchinson, line 53: `maxDataBytes` is a little more precise (if the `Update` method was called `Write` I would have let this slide).

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+		if err := br.dev.ReadBlocks(br.pos/br.dev.BlockSize(), br.buf); err != nil {
+			return 0, err
+		}
+	}
+	l := copy(b, br.buf[br.pos%br.dev.BlockSize():])
+	br.pos += uint(l)
+	return l, nil
+}
+
+func padToBlock(in []byte, blockSize uint) []byte {
```
mhutchinson, line 252: Is this necessary here? It looks like you already to the padding at a lower level?

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+		e, err := unmarshalEntry(br)
+		if err != nil {
+			if lastEntry != nil {
+				// We already found the lastet record in the journal, so we're done.
+				j.current = *lastEntry
+				j.nextBlock = lba
+				return nil
+			}
+			// Since we couldn't unmarshal the entry struct we'll have to fall back to
+			// scanning all blocks.
```
mhutchinson, line 149: It might be worth extending this comment to be clear about the case this captures. It looks quite generic, but I believe it's quite a special case actually: on the last journal write, it failed during the write, and this write was a wraparound that started at the beginning. In all other cases, either a) there is no data to read (because nothing was ever written), or b) `lastEntry != nil`.

While reading this I went "yeah that makes sense", then "hold on, no", "oh yeah, in case any data got corrupted at rest", "wait a sec, we don't tolerate that. This can never happen" *lightbulb moment* "right, the first element but only after a wraparound".

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if err := j.init(); err != nil {
+		return nil, err
+	}
+
+	return j, nil
+}
+
+// Data returns the application data from the most recent valid entry in the
+// journal, along with the entry's revision number.
```
mhutchinson, line 99: If this returns revision=0 then no data was ever written?

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+			continue
+		} else if e.Revision < lastEntry.Revision {
+			// We've found an older revision following a newer one, so we're done
+			j.current = *lastEntry
+			return nil
+		} else {
+			return fmt.Errorf("journal is corrupt - found two entries with the same revision (%d)", e.Revision)
+		}
+	}
+	return nil
```
mhutchinson, line 165: I think you need to do something here with assigning `lastEntry` to `j.current` ? There is a case where you update `lastEntry` above, and if that perfectly aligned with the end of the journal then it looks like you'll then snap out of the loop and then drop it on the floor.

It would also be good to be explicit about how `j.current` is initialized in the case of an empty journal. I hinted at it on the comment for `Data()` above, but it's an important case to make sure we get right.

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/trillian-examples/witness/golang/omniwitness/usbarmory/storage/testonly"
+)
+
+func magic0Hdr() [4]byte {
+	return [4]byte{magic0[0], magic0[1], magic0[2], magic0[3]}
+}
+
+func TestEntryMarshal(t *testing.T) {
```
mhutchinson, line 18: TODO: review these tests.

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
AlCutter: Added a bunch of commentary about that.

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
AlCutter: That would likely be bad, if `init` fails, then calling any of the methods on `j` is probably not what you want to do (if it's screwed and you just want to give up and start over, then explicitly using the raw access to zero out the blocks is probably the right thing to do, at least for now).

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
AlCutter: Good spot, ta

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
AlCutter: Good catch, I shifted that around a bit - removed this now.

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	if j.length != length {
+		t.Fatalf("Journal.length=%d, want %d", j.length, length)
+	}
+}
+
+func TestWriteSizeLimit(t *testing.T) {
+	storageBlocks := uint(20)
+	md := testonly.NewMemDev(t, storageBlocks)
+	start, length := uint(1), uint(storageBlocks)
```
mhutchinson, line 159: The cast of `storageBlocks` to a `uint` seems redundant?

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
```diff
@@ -0,0 +1,231 @@
...
+	limit := int((storageBlocks * md.BlockSize() / 3) - entryHeaderSize)
+	if err := j.Update(fill(limit, "ok...")); err != nil {
+		t.Fatalf("Update: %v, but expected write to succeed", err)
+	}
+	if err := j.Update(fill(limit+1, "BOOM")); err == nil {
+		t.Fatal("Update succeeded, but expected write fail")
+	}
+}
+
+func TestRoundTrip(t *testing.T) {
```
mhutchinson, line 175: Would be great to add a test that covers the case mentioned above where one of the writes perfectly fit byte-wise into the end of storage. I think that would be the first of the `doing` below, but then you don't easily get to test the wrap-around.

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
AlCutter: Thanks - I've expanded the comment, does that capture it better?

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return j.dev.WriteBlocks(j.nextBlock, padToBlock(buf.Bytes(), j.dev.BlockSize()))
+}
+
+// Init scans the journal to figure out the latest valid record, if any.
+func (j *Journal) init() error {
+	j.nextBlock = j.start
+	// Start where all good stories do: at the beginning!
+	lba := j.start
+	var lastEntry *entry
```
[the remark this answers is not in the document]
AlCutter: Done

witness/golang/omniwitness/usbarmory/storage/slots/journal.go
AlCutter: I _think_ it was ok, but I've jiggered thing around a bit to try to make it clearer - now we only assign to j.current etc. in one place (at the bottom of the func) - early returns became `break`s etc.

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
```diff
@@ -0,0 +1,231 @@
...
+				DataSHA256: sha256.Sum256([]byte("hello")),
+				Data:       []byte("hello"),
+			},
+			wantErr: true,
+		}, {
+			name: "bad hash",
+			e: &entry{
+				Magic:      magic0Hdr(),
+				Revision:   42,
+				DataLen:    uint64(len("hello")),
```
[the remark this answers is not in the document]
AlCutter: I'm not sure what the expected result should be, though.
e.g. it _could_ say _"this record, in this journal geometry, should not have been allowed to have been written since it's too large"_, but do we want that behaviour? 
(Actually, it couldn't because `unmarshalEntry` doesn't have access to the journal geometry, but the same question about behaviour is true for any arbitrary upper limit on allowable entry data size.)

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+func TestOpenJournal(t *testing.T) {
+	md := testonly.NewMemDev(t, 1)
+	start, length := uint(1), uint(len(md)-2)
+	j, err := OpenJournal(md, start, length)
+	if err != nil {
+		t.Fatalf("OpenJournal: %v", err)
+	}
+	if j.start != start {
+		t.Fatalf("Journal.start=%d, want %d", j.start, start)
```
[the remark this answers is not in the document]
AlCutter: done

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
AlCutter: Removed

witness/golang/omniwitness/usbarmory/storage/slots/journal_test.go
AlCutter: hehe - I added such a test earlier :)

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
```diff
@@ -0,0 +1,295 @@
...
+			//  b) the previously good entry/ies at the start of the journal
+			//     have been completely or partially overwritten during a
+			//     failed write attempt.
+			// Either way, we don't have a valid entry wth a length field we can
+			// rely on, so we we'll have to fall back to scanning all blocks to
+			// look for one.
+			lba++
+			continue
+		}
+		if lastEntry.Revision == 0 || e.Revision > lastEntry.Revision {
```
mhutchinson, line 183: I suspect `lastEntry.Revision == 0` is redundant, because if it is 0 then the next check should trivially pass (because revision is uint, and real revisions are always positive).

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
```diff
@@ -0,0 +1,295 @@
...
+		nextWriteLBA = j.start
+	}
+	j.nextBlock = nextWriteLBA
+	j.current = lastEntry
+
+	return nil
+}
+
+// unmarshalEntry reads and deserialises an entry structure from the provided reader.
+func unmarshalEntry(r io.Reader) (*entry, error) {
```
mhutchinson, line 215: Is there a reason to return a pointer here? Similarly with accepting it below.

I thought the go preference was generally to copy structs unless there's a good reason not to. But you've been in this game longer than me.

mhutchinson (approved):

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
AlCutter: Yeah, it was a mechanical change to replace the `nil` check with `Rev==0`, as you say there is a material difference in that previously a written entry which had a Revision of 0 would have been accepted, but in practice that should never happen so we can safely simplify this.

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
AlCutter: The main reason is just to not return a entry in some unknown/invalid state in the case of error unless that's a specific requirement (which would warrant a comment on the func about this behaviour) - this is mostly the pattern followed by the Go std lib (a notable exception is probably the `Reader` types which can return valid data and error, but that's because it can return an EOF "error").

PR #619, opened 2022-03-28
Use the context passed in

mhutchinson:
This will cause this feeder to bail when the context becomes cancelled. I'm surprised there isn't a lint check for calling one of the methods like context.Background when there is already a ctx in scope.


AlCutter (approved):
Good catch!

PR #622, opened 2022-03-29
Unikernel omniwitness can now distribute

mhutchinson:
If the github authentication details are provided then the distributors
will be started.

This also adds  goroutine start/stop info logging, which can be
useful in case there are any goroutines that exit early, and any that
don't automatically exit when the context is cancelled.

### DIFF
--- a/witness/golang/omniwitness/internal/omniwitness/omniwitness.go
+++ b/witness/golang/omniwitness/internal/omniwitness/omniwitness.go
@@ -158,7 +158,7 @@ func Main(ctx context.Context, operatorConfig OperatorConfig, httpListener net.L
 		})
 	}
 
-	if len(operatorConfig.GithubUser) > 0 {
+	if operatorConfig.GithubUser != "" {
 		var distLogs = []dist_gh.Log{}
 		for config := range feeders {
 			// TODO(mhutchinson): This verifier could be created inside the distributor.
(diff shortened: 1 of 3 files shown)


witness/golang/omniwitness/internal/omniwitness/omniwitness.go
```diff
@@ -148,13 +182,93 @@ func Main(ctx context.Context, signer note.Signer, httpListener net.Listener, ht
...
+			Repo:            repo,
+			DistributorPath: ".",
+			Logs:            logs,
+			WitSigV:         operatorConfig.WitnessVerifier,
+			Witness:         witness,
+		}
+
+		return distribute(opts)
+	})
+}
```
AlCutter, line 270: Could this be factored out and shared between the two distributor goroutines?

witness/golang/omniwitness/internal/omniwitness/omniwitness.go
```diff
@@ -135,11 +152,28 @@ func Main(ctx context.Context, signer note.Signer, httpListener net.Listener, ht
...
 		// Continually feed this log in its own goroutine, hooked up to the global waitgroup.
 		g.Go(func() error {
+			glog.Infof("Feeder %q goroutine started", c.Origin)
+			defer glog.Infof("Feeder %q goroutine done", c.Origin)
 			return f(ctx, c, bw, httpClient, feedInterval)
 		})
 	}
 
-	// TODO(mhutchinson): Start the distributors too if auth details are present.
+	if len(operatorConfig.GithubUser) > 0 {
```
AlCutter, line 161: nit: I think go style might prefer `if opCfg.GitHubUser != ""` since this `len` is really checking for empty/not-empty

witness/golang/omniwitness/internal/omniwitness/omniwitness.go
```diff
@@ -67,11 +72,23 @@ type multiLogFeederConfig struct {
 	Logs []config.Log `yaml:"Logs"`
 }
 
+// OperatorConfig allows the bare minimum operator-specific configuration.
+// This should only contain configuration details that are custom per-operator.
+type OperatorConfig struct {
+	WitnessSigner   note.Signer
+	WitnessVerifier note.Verifier
+
+	GithubUser  string
```
AlCutter, line 81: nit: strictly, I think it should be GitHub :)

witness/golang/omniwitness/internal/omniwitness/omniwitness.go
mhutchinson: When writing this initially I created a top-level function to do this and it just felt like a giant handbag that took as many arguments as it had lines of code within it. I've extracted another function in this function to create the configuration, which I think is a reasonable balance. Having it as a nested function allows it to capture all of the args to `runDistributors` that it also needs, so saves redeclaring all of that.

witness/golang/omniwitness/internal/omniwitness/omniwitness.go
mhutchinson: Yeah, but this is way nicer to read, and easier to identify what the interesting part is than the strictly correct version. Honestly, I find `GitHubUser` a little distressing on its own, and when followed by 2 other similarly named fields it's just too much.

witness/golang/omniwitness/internal/omniwitness/omniwitness.go
mhutchinson: Done.

witness/golang/omniwitness/internal/omniwitness/omniwitness.go
mhutchinson: Discussed offline. Resolving this.

AlCutter (approved):

PR #627, opened 2022-03-31
Remove unused Cloud Build substitution added from https://github.com/google/trillian-examples/pull/598.

jiggoha:
The Cloud Build workflow currently fails with the following error:

```
Your build failed to run: generic::invalid_argument: generic::invalid_argument: invalid value for 'build.substitutions': key in the template "ENTRIES_DIR" is not a valid built-in substitution
```

### DIFF
--- a/cloudbuild_docker.yaml
+++ b/cloudbuild_docker.yaml
@@ -153,4 +153,3 @@ steps:
 substitutions:
   _DOCKER_BUILDX_PLATFORMS: 'linux/amd64,linux/arm/v7'
   _BUCKET: 'transparency_log'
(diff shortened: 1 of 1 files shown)


AlCutter (approved):

mhutchinson:
Weirdly I just found and fixed this via another entry point. But I think this is entirely the wrong fix. The problem is that this declares the variable,  but when it's used it doesn't have the underscore. https://github.com/google/trillian-examples/pull/629 is the right fix, I think?

jiggoha:
Closing this since Martin wrote https://github.com/google/trillian-examples/pull/629.

PR #629, opened 2022-03-31
Fix cloudbuild_docker

mhutchinson:
Error in triggered runs is 'Your build failed to run: generic::invalid_argument: generic::invalid_argument: invalid value for 'build.substitutions': key in the template ENTRIES_DIR is not a valid built-in substitution'.

### DIFF
--- a/cloudbuild_docker.yaml
+++ b/cloudbuild_docker.yaml
@@ -74,7 +74,7 @@ steps:
       - |
         echo "bucket ${_BUCKET}"
         echo "log entry content ${REPO_NAME},${COMMIT_SHA},$(cat /workspace/image-digest.txt)\"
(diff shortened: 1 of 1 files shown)


mhutchinson:
Are `REPO_NAME` and `COMMIT_SHA` built-ins? If not, it will fail there after this is applied.

mhutchinson:
> Are `REPO_NAME` and `COMMIT_SHA` built-ins? If not, it will fail there after this is applied.

Yes: https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values

mhutchinson:
This is attempting to fix the same problem as #627 which I hadn't noticed when I sent this out.

jiggoha (approved):

PR #630, opened 2022-03-31
[Witness/USBArmory] Use DCP for hashing

AlCutter:


witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
```diff
@@ -102,12 +106,13 @@ func (e *entry) Size() int {
 // OpenJournal returns a new journal structure for interacting with a journal stored in the
 // [start, start+length) range of blocks accessible via dev.
 // Journal ranges should not overlap with one another, or corruption will almost certainly occur.
-func OpenJournal(dev BlockReaderWriter, start, length uint) (*Journal, error) {
+func OpenJournal(dev BlockReaderWriter, start, length uint, h Hasher) (*Journal, error) {
```
mhutchinson, line 109: It's obvious, but I think worth noting in the comment that the hasher passed in must functionally match any previous hasher used in this journal, or else everything will look corrupted!

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
```diff
@@ -43,11 +43,15 @@ type BlockReaderWriter interface {
 	WriteBlocks(lba uint, b []byte) error
 }
 
+// Hasher returns a hash of the data available from the given reader.
+type Hasher func(io.Reader) ([32]byte, error)
```
mhutchinson, line 47: Any reason to take an `io.Reader` ? It looks like there's always the `[]byte` read into memory when we do the hashing.

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
```diff
@@ -102,12 +106,13 @@ func (e *entry) Size() int {
...
 // [start, start+length) range of blocks accessible via dev.
 // Journal ranges should not overlap with one another, or corruption will almost certainly occur.
-func OpenJournal(dev BlockReaderWriter, start, length uint) (*Journal, error) {
+func OpenJournal(dev BlockReaderWriter, start, length uint, h Hasher) (*Journal, error) {
 	j := &Journal{
 		dev:          dev,
 		start:        start,
 		length:       length,
 		maxDataBytes: (length*dev.BlockSize())/minEntries - entryHeaderSize,
+		hasher:       h,
```
mhutchinson, line 115: nit: probably subjective style, but while I prefer `hasher` for the field name from a lifetime of doing OO, from an applied perspective, `hash` is a better name because it's a straight function. I think `hash` looks better at the point of use.

witness/golang/omniwitness/usbarmory/main.go
```diff
@@ -64,9 +67,36 @@ func init() {
...
+	}
+	bs := 32 << 10
+	buf := make([]byte, bs)
+	var n int
+	for err == nil {
+		n, err = r.Read(buf)
+		if n > 0 {
+			h.Write(buf[:n])
+		}
+
```
mhutchinson, line 95: nit: blank line considered confusing :-)

AlCutter: that was to add a delay each time around the loop :)

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
AlCutter: renamed :)

witness/golang/omniwitness/usbarmory/internal/storage/slots/journal.go
AlCutter: I've just renamed to make it clear this is always just SHA256 (but allowing for different implementations of it)

mhutchinson: Hmmm, I guess that works. Any validation scheme would work as long as it was consistent, but if you nail it down to sha256 then it's always consistent.

witness/golang/omniwitness/usbarmory/main.go
mhutchinson: LGTM.

mhutchinson (approved):

PR #631, opened 2022-03-31
[Witness/USBArmory] Lazy-open slot journals

AlCutter:
There is some work required when opening journals, so only attempt to open them when the app has asked for it.


witness/golang/omniwitness/usbarmory/internal/storage/slots/slots.go
```diff
@@ -116,6 +117,14 @@ func (p *Partition) Open(slot uint) (*Slot, error) {
...
 	}
 	s := &p.slots[slot]
+	if s.journal == nil {
+		glog.V(2).Infof("Opening journal for slot %d", slot)
+		j, err := OpenJournal(p.dev, s.start, s.length, p.sha256)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open journal for slot %d: %v", slot, err)
+		}
+		s.journal = j
+	}
```
mhutchinson, line 127: I _think_ this should be an `EnsureInit` method on `Slot`, and it should be guarded by the lock there. As this stands, unless we know that this `Open` is only called from a single thread, multiple journals could be opened for the same slot. Obviously only the last one wins, but I suspect it's a source of hard to find bugs and not how we reason the initialization should work.

AlCutter: Yeah, good spot - this is better, ta.

witness/golang/omniwitness/usbarmory/internal/storage/slots/slots.go
```diff
@@ -117,14 +117,11 @@ func (p *Partition) Open(slot uint) (*Slot, error) {
...
-	if s.journal == nil {
-		glog.V(2).Infof("Opening journal for slot %d", slot)
-		j, err := OpenJournal(p.dev, s.start, s.length, p.sha256)
-		if err != nil {
-			return nil, fmt.Errorf("failed to open journal for slot %d: %v", slot, err)
-		}
-		s.journal = j
+	glog.V(2).Infof("Opening slot %d", slot)
+	if err := s.Open(p.dev, p.sha256); err != nil {
+		glog.V(2).Infof("Failed to open slot %d: %v", slot, err)
```
mhutchinson, line 122: This should probably return the error too?

mhutchinson (approved):
Please address the comment, but pre-emptive approval.

witness/golang/omniwitness/usbarmory/internal/storage/slots/slots.go
AlCutter: The method comment does cover that case already :)

PR #637, opened 2022-04-05
Set up the docker build using the same setup as the others

mhutchinson:
The previous dockerfile wouldn't build in GCB. This is failing for weird reasons like TLS timeouts. Using the same base image and setup as the witness Dockerfile should make this work. At least, this works on my dev machinne.


AlCutter (approved):
Mmokay!

PR #655, opened 2022-05-27
Upgrade lint to more recent version and fixed errors

mhutchinson:
This started off as trying to make the lint configuration consistent with other repos, and removing `skip` instruction that no longer matched anything. In doing so, it turned out that we were using an old version, and updating that caused a ream of lint warnings. So I've fixed them.

This also brought to light some weirdness where we install multiple versions of go (because we install tamago). I've made it so that:
 a) the same version of both go and tamago is installed
 b) tamago doesn't blat go in the path; invoking `go` will do normal go, not tamago

### DIFF
--- a/sumdbaudit/client/sumdb_test.go
+++ b/sumdbaudit/client/sumdb_test.go
@@ -67,7 +67,7 @@ func TestLeavesAtOffset(t *testing.T) {
 			t.Errorf("expected string terminating in newline, got: %x", l)
 		}
 		expStart := "golang.org/x/"
-		if got, want := fmt.Sprintf("%s", l[:len(expStart)]), expStart; got != want {
+		if got, want := string(l[:len(expStart)]), expStart; got != want {
 			t.Errorf("got prefix '%s', wanted '%s'", got, want)
 		}
 	}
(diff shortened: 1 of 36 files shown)


mhutchinson:
I've noticed some other stuff out of date here. Going to use this as an opportunity to fix up the lint across the repo and bring this up to a better standard.

sumdbaudit/client/sumdb_test.go
```diff
@@ -67,7 +67,7 @@ func TestLeavesAtOffset(t *testing.T) {
 			t.Errorf("expected string terminating in newline, got: %x", l)
 		}
 		expStart := "golang.org/x/"
-		if got, want := fmt.Sprintf("%s", l[:len(expStart)]), expStart; got != want {
+		if got, want := string(l[:len(expStart)]), expStart; got != want {
```
AlCutter, line 70: Does this need the `string()` cast?

AlCutter (approved):
Thanks for doing this!

One minor question inline.

sumdbaudit/client/sumdb_test.go
mhutchinson: Yes, because `want` is a string, and `l` is `[]byte`. You can't `!=` `string` and `[]byte`.

PR #672, opened 2022-08-19
Conditionally write sequence number. 

jiggoha:
If it already exists, try the next number.

See documentation:
  * https://cloud.google.com/storage/docs/object-versioning
  * https://cloud.google.com/storage/docs/metadata#generation-number
  * https://cloud.google.com/storage/docs/request-preconditions#special-case

### DIFF
--- a/serverless/experimental/gcp-log/internal/storage/storage.go
+++ b/serverless/experimental/gcp-log/internal/storage/storage.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


serverless/experimental/gcp-log/internal/storage/storage.go
```diff
@@ -268,14 +271,30 @@ func (c *Client) Sequence(ctx context.Context, leafhash []byte, leaf []byte) (ui
 		}
 
 		// Found the next available sequence number; write it.
-		w := bkt.Object(seqPath).NewWriter(ctx)
+		//
+		// Conditionally write only if the object does not exist yet:
+		// https://cloud.google.com/storage/docs/request-preconditions#special-case.
+		// This may exist if there is more than one instance of the sequencer
+		// writing to the same log.
+		w := bkt.Object(seqPath).If(gcs.Conditions{DoesNotExist: true}).NewWriter(ctx)
```
AlCutter, line 279: Would be good to know when this precondition is applied, I couldn't really tell from the docs :/

Hopefully it's at `Write()` or `Close()` time otherwise I think this is still racy (unless it does something clever like having `NewWriter` take a lease).

AlCutter (approved):

PR #681, opened 2022-10-05
Fix Rekor feeder to handle sharded logs

haydentherapper:
This updates the feeder to specify the shard ID when fetching a consistency proof, using the ID from the config.

I also needed to update fetching the latest checkpoint, searching for the checkpoint that matches the shard ID from the config.

Also added the new shard to the configs, and updated the Origin to match what's now returned by Rekor.

Signed-off-by: Hayden Blauzvern <hblauzvern@google.com>

### DIFF
--- a/internal/feeder/rekor/rekor_feeder.go
+++ b/internal/feeder/rekor/rekor_feeder.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -76,6 +75,8 @@ func FeedLog(ctx context.Context, l config.Log, w feeder.Witness, c *http.Client
 	}
 
 	fetchCP := func(ctx context.Context) ([]byte, error) {
+		// Each Rekor feeder will request the same log info.
+		// TODO: Explore if it's feasible to request this once for all Rekor feeders.
 		li := logInfo{}
 		if err := getJSON(ctx, c, lURL, "api/v1/log", &li); err != nil {
 			return nil, fmt.Errorf("failed to fetch log info: %v", err)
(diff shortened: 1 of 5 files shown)


haydentherapper:
@AlCutter, what's the best way to test this? I ran the feeder locally without a witness running, and the feeder successfully fetched multiple checkpoints and verified them.

internal/feeder/rekor/rekor_feeder.go
```diff
@@ -65,14 +80,24 @@ func FeedLog(ctx context.Context, l config.Log, w feeder.Witness, c *http.Client
 		if err := getJSON(ctx, c, lURL, "api/v1/log", &li); err != nil {
 			return nil, fmt.Errorf("failed to fetch log info: %v", err)
 		}
-		return []byte(li.SignedTreeHead), nil
+		// Active shard
```
AlCutter, line 83: Not a blocker, but I think this means that for every shard in the config, we're going to re-request the same JSON from the loginfo endpoint and then pull out just the bit for the relevant shard.

Might be worth popping a comment to this effect, or perhaps a `TODO` to investigate (at some point in the future) whether we can, or even should, improve the efficiency by fetching once per sigstore instance?
It may very well not be worth the lift in complexity, or the special-casing which would be necessary for the config, etc. but having the note in there will help future us when we come across this again :)

internal/feeder/rekor/rekor_feeder.go
```diff
@@ -65,14 +80,24 @@ func FeedLog(ctx context.Context, l config.Log, w feeder.Witness, c *http.Client
...
+		if li.TreeID == l.ID {
+			return []byte(li.SignedTreeHead), nil
+		}
+		// Search inactive shards
+		for _, shard := range li.InactiveShards {
+			if shard.TreeID == l.ID {
+				return []byte(shard.SignedTreeHead), nil
+			}
+		}
+		return nil, errors.New("failed to find shard that matched log ID from config")
```
AlCutter, line 93: super nit: I'd be tempted to do the `fmt.Errorf(... l.ID)` thing here; seeing what's being looked for could help take one indirection off trying debug an issue when it's not working.

internal/feeder/rekor/rekor_feeder.go
haydentherapper: Added a comment. Another option would be to design this such that it automatically pulls in every shard - This is something I'm exploring in sigstore/rekor-monitor.

haydentherapper:
@AlCutter Updated! Once this is merged, I'll send out PRs to update the distributors https://github.com/mhutchinson/mhutchinson-distributor and https://github.com/WolseyBankWitness/rediffusion. Let me know if you know of any other places that need to be updated.

AlCutter (approved):

haydentherapper:
Thanks, can you merge?

PR #683, opened 2022-10-07
[Witness/Armory] Add support for DHCP

AlCutter:
Uses (parts of) the Fuscia DHCP implementation to add support for configuring the armory network interface via DHCP4.

We copy over the relevant DHCP code rather than take a dep because a) the fuscia repo is huge, and b) depends upon an older (and incompatible) version of gVisor than the one this repo uses, so needed some small fixups.


witness/golang/omniwitness/usbarmory/usbnet.go
```diff
@@ -49,6 +64,12 @@ func initNetworking() error {
 	imx6ul.USB1.Init()
 	imx6ul.USB1.DeviceMode()
 	imx6ul.USB1.Reset()
+
+	if ip == "" {
+		netReady.Add(1)
+		glog.Info("NIC has no address, using DHCP4")
```
phbnf, line 70: Nit: "no IP address yet" maybe?

phbnf (approved):

witness/golang/omniwitness/usbarmory/usbnet.go
AlCutter: Done, ta!

PR #689, opened 2022-11-03
Optionally output log inclusion proof in JSON.

arai-fortanix:
For some scenarios, it is useful to independently check a log inclusion proof. The serverless "client inclusion" command does provide the information necessary to verify the proof, such as the node index, the tree size, the tree root hash, and node hashes. But these are not provided in an easy-to-consume format. The hashes can be written to a file, but they are useless without the node index, tree size, and root hash. The index, tree size and root hash have to be scraped from the log output, which requires messy text parsing.

This change adds a new option to the "client inclusion" command to write all of this information to a file in JSON format. This gives us a self-contained proof file that can be used to verify log inclusion.

### DIFF
--- a/clone/cmd/serverlessclone/serverlessclone.go
+++ b/clone/cmd/serverlessclone/serverlessclone.go
@@ -82,7 +82,7 @@ func main() {
 	}
 	f := newFetcher(u)
 
-	targetCp, rawCp, err := client.FetchCheckpoint(ctx, f, v, *origin)
+	targetCp, rawCp, _, err := client.FetchCheckpoint(ctx, f, v, *origin)
 	if err != nil {
 		glog.Exitf("Failed to get latest checkpoint from log: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


AlCutter:
The `client` command was intended to be an example of how to interact with a serverless log which might then inspire folks to create their own client implementations that work just how they need them to - e.g. either by directly embedding log client code in the thing which needs to verify proof of inclusion, or if that's not possible/desired, then perhaps writing out proof bundles in the most appropriate encoding for the particular use-case.
An example of this is here: https://github.com/usbarmory/armory-drive-log/blob/master/cmd/create_proofbundle/main.go

Adding such a JSON output to this example command would be adding a tacit "hidden" API surface to it which I think I'd rather avoid at the moment (the current outputs were intended for debugging - we should probably remove those!)

arai-fortanix:
Fair enough. We can write our own tool to do this. I think we still need the change to FetchCheckpoint that returns the note so external code can verify the signature itself. I can submit a new pull request with just that change.

PR #690, opened 2022-11-04
[Witness/Feeder] Configure Rekor treeID via URL

AlCutter:
This PR moves the configuration of the Rekor `treeID` into the `LogURL` field rather than overloading `ID`.

The URL is a more natural place for this param (indeed, `treeID` ends up as a HTTP param on the request), and overriding `ID` has the unfortunate side effect of changing the directory name for the log in current distributor implementations away from the default derived hash value to a value which all distributor clients would need to configure.

### DIFF
--- a/internal/feeder/rekor/rekor_feeder.go
+++ b/internal/feeder/rekor/rekor_feeder.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -64,10 +65,17 @@ type proof struct {
 // FeedLog feeds checkpoints from the source log to the witness.
 // If interval is non-zero, this function will return when the context is done, otherwise it will perform
 // one feed cycle and return.
+//
+// Note that this feeder expects the configured URL to contain a "treeID" query parameter which contains the
+// correct Rekor log tree ID.
 func FeedLog(ctx context.Context, l config.Log, w feeder.Witness, c *http.Client, interval time.Duration) error {
 	lURL, err := url.Parse(l.URL)
 	if err != nil {
 		return fmt.Errorf("invalid LogURL %q: %v", l.URL, err)
+	}
+	treeID := lURL.Query().Get("treeID")
+	if treeID == "" {
+		return errors.New("configured LogURL does not contain the required treeID query parameter")
 	}
 	logSigV, err := i_note.NewVerifier(l.PublicKeyType, l.PublicKey)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


AlCutter:
@haydentherapper wdyt?

AlCutter:
@haydentherapper - just floating this up your inbox again :)

haydentherapper:
Sorry! I will take a look asap at this.

witness/golang/omniwitness/distributor_configs/witness.yaml
```diff
@@ -8,13 +8,11 @@ Logs:
   - Origin: rekor.sigstore.dev - 3904496407287907110
     PublicKeyType: ecdsa
     PublicKey: rekor.sigstore.dev+c0d23d6a+AjBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABNhtmPtrWm3U1eQXBogSMdGvXwBcK5AW5i0hrZLOC96l+smGNM7nwZ4QvFK/4sueRoVj//QP22Ni4Qt9DPfkWLc=
-    URL: https://rekor.sigstore.dev
-    ID: 3904496407287907110
+    URL: https://rekor.sigstore.dev?treeID=3904496407287907110
```
haydentherapper, line 11: Would it be better to add another configuration field rather than overload URL?

internal/feeder/rekor/rekor_feeder.go
```diff
@@ -98,7 +106,7 @@ func FeedLog(ctx context.Context, l config.Log, w feeder.Witness, c *http.Client
 			return [][]byte{}, nil
 		}
 		cp := proof{}
-		if err := getJSON(ctx, c, lURL, fmt.Sprintf("api/v1/log/proof?firstSize=%d&lastSize=%d&treeID=%s", from.Size, to.Size, l.ID), &cp); err != nil {
+		if err := getJSON(ctx, c, lURL, fmt.Sprintf("api/v1/log/proof?firstSize=%d&lastSize=%d&treeID=%s", from.Size, to.Size, treeID), &cp); err != nil {
```
haydentherapper, line 109: If we go with this approach, do we also need to remove the query parameter from `lURL`?

witness/golang/omniwitness/distributor_configs/witness.yaml
AlCutter: My thinking was that the ID is (at least logically) part of the URL - if the log were exposed as a REST API it'd be something like `rekor.example.com/log/<treeID>/get-root` etc., just here it's being sent as a GET param...

internal/feeder/rekor/rekor_feeder.go
haydentherapper: Sorry for the delay - Awesome, that's good! This approach LGTM then.

haydentherapper (approved):

phbnf (approved):

PR #697, opened 2022-11-21
Upgrade Go version from 1.17 to 1.19

roger2hk:


AlCutter (approved):
Big lift, thanks for doing this Roger!

PR #700, opened 2022-11-29
Make checkpoint signatures available from checkpoint tracker.

arai-fortanix:
This updates the transparency log checkpoint tracker to make the checkpoint's signature(s) available for client code that wants to verify the signature itself.

### DIFF
--- a/serverless/integration/integration_test.go
+++ b/serverless/integration/integration_test.go
@@ -24,8 +24,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"strconv"
-	"strings"
 	"testing"
 
 	"github.com/golang/glog"
@@ -74,6 +72,7 @@ func RunIntegration(t *testing.T, s log.Storage, f client.Fetcher, lh *rfc6962.H
 		// Sequence some leaves:
 		leaves := sequenceNLeaves(ctx, t, s, lh, i*leavesPerLoop, leavesPerLoop)
 
+		var latestCpNote *note.Note = nil
 		// Integrate those leaves
 		{
 			update, err := log.Integrate(ctx, checkpoint, s, lh)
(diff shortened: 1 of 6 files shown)


roger2hk:
```
Already have image: gcr.io/trillian-opensource-ci/testbase
+ ./scripts/presubmit.sh --no-build --no-linters --no-generate --no-docker --no-actions
Building serverless wasm ====================
# github.com/google/trillian-examples/serverless/experimental/wasm
serverless/experimental/wasm/main.go:92:17: assignment mismatch: 3 variables but client.FetchCheckpoint returns 4 values
serverless/experimental/wasm/main.go:248:17: assignment mismatch: 3 variables but client.FetchCheckpoint returns 4 values
```

serverless/integration/integration_test.go
```diff
@@ -93,6 +96,32 @@ func RunIntegration(t *testing.T, s log.Storage, f client.Fetcher, lh *rfc6962.H
 			t.Fatalf("Failed to update tracked log state: %q", err)
 		}
 		newCheckpoint := lst.LatestConsistent
+		// Verify that the returned checkpoint note is as expected.
```
AlCutter, line 99: Can you just save the `cpNote` (and perhaps `cpNoteSigned`) from L84/L85 and compare them with what you get back from `Update`?

It seems like a lot of the testing below is recovering things which should already be covered by tests in the `log` package (around the correct form of a `log.Checkpoint`, etc.)

AlCutter (commented):
Also, it looks like there might be more references to the updated function to fix - see Roger's comment above.

AlCutter (commented):
Sorry for the slow response, just one little nit inline.

AlCutter (approved):

PR #712, opened 2023-01-10
cmd/addrwitness: a new addressable or bastion-mediated witness

FiloSottile:
This uses `internal/witness` like `cmd/witness`, but with a few major differences:

* in general, addrwitness is designed to be directly addressable from the Internet, without a fetcher, enabling low-latency logs to fetch co-signatures synchronously and distribute them along with their tree heads
* optionally, addrwitness supports being exposed through a [bastion](https://git.glasklar.is/sigsum/project/documentation/-/blob/main/bastion.md), for witnesses that are not easily exposed to the Internet
* the witness adds two signatures: a plain one for backwards compatibility, and a timestamped one
* the get API returns the tree size of the latest checkpoint, not the whole checkpoint, to prevent misuse and direct logs to take care of co-signature distribution
* logs are indexed by the origin string, and there is no concept of a separate log ID
  * this is important and arguably a security fix: see the comments in `main.go` or [this](https://git.glasklar.is/sigsum/project/documentation/-/blob/6e3a680db9b808b536921978526ed3f5e069153f/archive/2021-08-10-witnessing-broader-discuss#L95-115)

This PR is a proof of concept. In particular, it doesn't support compact ranges or ECDSA signatures, and has no tests or metrics, and its handling of reconnection to the bastion is clunky.


AlCutter:
It would be nice if this could be made to fit inside a `feeder` implementation as it would then Just Work™️ inside the omniwitness etc.

e.g. the [sumdb feeder](https://github.com/google/trillian-examples/blob/master/internal/feeder/sumdb/sumdb_feeder.go) - the `fetchCheckpoint` and `fetchProof` funcs could poll the bastion (e.g. `getCheckpoint` polls for CP+poof, returns CP and stores the proof to be returned by the next call to `fetchProof` or something?)

Might need a bit of refactoring or some sprinkles on the configuration front...

FiloSottile:
(Didn't realize cleaning up forks would close PRs, it didn't use to be the case, but I'd say we came a long way since this, so there is indeed no need to keep this except for historical reference.)

PR #714, opened 2023-01-12
Made the monolith easier to discover, and better documented

mhutchinson:
The monolith should be the default witness that most people should find themselves at; it is simple to deploy and follows all known logs. The other configurations (dockerized, bare witness) are still useful, but only to advanced users that know they don't want the monolith.


AlCutter (approved):
Nice one.

We could also consider tweaking the directory structure a bit too; moving `monolith` under `witness/golang/cmd`, and perhaps moving the individual binaries to something like `witness/golang/cmd/{advanced,components,???}/...`?

PR #734, opened 2023-01-31
Added instructions for running monolith in docker

mhutchinson:


witness/golang/omniwitness/monolith/README.md
```diff
@@ -6,7 +6,7 @@ common configuration steps such as key generation and creating github credential
 This version uses a single `main` construct to invoke all of the sub-components needed to
 implement the OmniWitness. 
 
-## Running
+## Running Bare
```
AlCutter, line 9: How about "locally" or "... the binary"?
I wouldn't know what "Bare" meant in this context until I'd read through the section

AlCutter (approved):
Just one small nit/suggestion on the section title.

witness/golang/omniwitness/monolith/README.md
mhutchinson: I renamed it to "running without docker" and moved it to after the docker section. I think that makes more sense?

PR #750, opened 2023-02-20
Enable all lint checks

mhutchinson:
Previously only named linters were ran, but the benefits of doing this are lost in time. Some of the linters that are excluded from this technique provide good value (such as checking that errors are inspected).

### DIFF
--- a/binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
+++ b/binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
@@ -48,7 +48,6 @@ func (ws *Storage) init() error {
 	if err != nil {
 		return fmt.Errorf("failed to open file: %w", err)
 	}
-	defer f.Close()
 	if err := f.Close(); err != nil {
 		return fmt.Errorf("failed to close file: %w", err)
 	}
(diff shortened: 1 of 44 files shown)


AlCutter (approved):
❤️

mhutchinson:
@AlCutter would be good to review this before it gets any conflicts. It's pretty wide.

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
```diff
@@ -64,12 +63,12 @@ func (ws *Storage) StoreCP(wcp []byte) error {
 	// Check if file exists, open for write and store the checkpoint
 	f, err := os.OpenFile(ws.fp, os.O_RDWR, fileMask)
 	if err != nil {
-		f.Close()
+		_ = f.Close()
```
AlCutter, line 66: Hmm, not your change, but this file-handling code seems a bit unusual in here, e.g.:
 - is `f` guaranteed to be non-nil if `err` is non-nil?
 - I wonder why we're doing `os.OpenFile`/`f.Write`/`f.Close` rather than `os.WriteFile`?
 
No need to fix in your PR, but we should probably make a note and look at this...

AlCutter (approved):

binary_transparency/firmware/cmd/ft_witness/internal/ws/store.go
mhutchinson: Addressed in #799

PR #774, opened 2023-04-04
Improve clone tool docker integration

mhutchinson:
This uses explicit Go versions for the builder, and uses alpine for both the builder image and the release image, which is much smaller.

There is now a docker-compose for sumdb that brings up a mariadb image and the sumdb cloner that will clone into it.


clone/cmd/sumdbclone/README.md
```diff
@@ -13,6 +13,34 @@ go run ./clone/cmd/sumdbclone --alsologtostderr --v=1 --mysql_uri 'clonetool:let
 
 ## Docker
 
+### SQL DB in Docker
```
AlCutter, line 16: Is it worth adding a pointer to the `rpi` docker-compose file? (or is there some magic in `docker-compose` which somehow does the right thing?)

clone/cmd/sumdbclone/docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
...
+      retries: 5
+
+  clone:
+    build: 
+      context: ../../..
+      dockerfile: ./clone/cmd/sumdbclone/Dockerfile
+    command: [
+      "--alsologtostderr",
+      "--v=2",
+      "--mysql_uri=sumdb:letmein@tcp(db:3306)/sumdb",
```
AlCutter, line 25: Is it possible to have common vars (or something) shared across both services so the DB username/password only needs to be set once?

AlCutter (approved):

clone/cmd/sumdbclone/README.md
mhutchinson: Updated this to use `docker compose` instead of `docker-compose` which uses a newer version (written in Go instead of Python), which supports the healthcheck syntax on rpi. It still doesn't do anything magical with the override file, so this is now documented explicitly.

clone/cmd/sumdbclone/docker-compose.yaml
mhutchinson: Pulled these things out to a .env file which does allow for variable expansion in other services.

mhutchinson:
Ta, @AlCutter. PTAL.

PR #777, opened 2023-04-05
Clean up sumdbaudit directory

mhutchinson:
Extracted sumdb feeder from sumdbaudit and moved it alongside the other feeders

Removed specific witness for sumdb; users should deploy the omniwitness unless there's a good reason not to. We don't want to be maintaining docker configs for these non-standard deployments though.


mhutchinson:
@AlCutter if you could take a look if you get a moment that would be ace. I think everything in this sumdbaudit package can be retired eventually. This was the PoC breeding ground for the cloner and the mirror and the witness, but all of this stuff now exists in a more generic form. The verifier is probably the only thing that is still unique, and that can be moved next to the cloner now to mirror the `ctverify` app.

AlCutter (approved):

PR #779, opened 2023-04-05
Use MariaDB instead of sqlite in tests

mhutchinson:
This uses docker to bring up a temporary database server, and each test creates a new database within it to prevent data collisions.

GCB is a bit of a problem in that it has some custom configuration required on the docker host. This problem is worked around by having a build tag for cloudbuild; when this is provided then the `ConfigureHost` function is defined to do this configuration, otherwise it does nothing.


AlCutter (approved):

distributor/cmd/internal/distributor/distributor_test.go
```diff
@@ -59,6 +66,98 @@ var (
...
+	nextDB  uint32
+}
+
+func (h *dbHelper) connect() (*sql.DB, error) {
+	connectString := fmt.Sprintf("root:secret@(%s)/mysql", h.address)
+	return sql.Open("mysql", connectString)
+}
+
+func (h *dbHelper) create(testName string) (*sql.DB, error) {
+	db, err := h.connect()
```
AlCutter, line 82: Should there be a `defer db.Close()` since we re-open the DB with a more specific URI below?

distributor/cmd/internal/distributor/distributor_test.go
```diff
@@ -59,6 +66,98 @@ var (
...
+func TestMain(m *testing.M) {
+	flag.Parse()
+	// uses a sensible default on windows (tcp/http) and linux/osx (socket)
+	pool, err := dockertest.NewPool("")
+	if err != nil {
+		glog.Fatalf("Could not construct pool: %s", err)
+	}
+
+	// uses pool to try to connect to Docker
+	err = pool.Client.Ping()
```
AlCutter, line 105: nit: this could collapse into the old `if err := ...; err != ...`

internal/testonly/docker/cloudbuild.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2023 Google LLC. All Rights Reserved.
```
AlCutter, line 1: Is it worth having a chew over build tags vs. an ENV var? Can chat offline if you like?

AlCutter (approved):

distributor/cmd/internal/distributor/distributor_test.go
mhutchinson: Done.

distributor/cmd/internal/distributor/distributor_test.go
mhutchinson: Done.

internal/testonly/docker/cloudbuild.go
mhutchinson: Yeah let's talk about this a bit.

AlCutter (approved):

PR #787, opened 2023-04-12
HTTP server and dockerization for distributor

mhutchinson:
Using the docker-compose scripts it is trivial to bring up a distributor instance along with a MariaDB instance to persist the state.


distributor/cmd/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM golang:1.19-alpine AS builder
+
+ARG GOFLAGS=""
+ENV GOFLAGS=$GOFLAGS
+ENV GO111MODULE=on
```
roger2hk, line 5: `GO111MODULE` is on by default since Go 1.16.

mhutchinson: Good call, now removed.

distributor/cmd/internal/http/server.go
```diff
@@ -0,0 +1,136 @@
...
+		return
+	}
+	w.Header().Set("Content-Type", "text/json")
+	w.Write(logList)
+}
+
+// RegisterHandlers registers HTTP handlers for witness endpoints.
+func (s *Server) RegisterHandlers(r *mux.Router) {
+	logStr := "{logid:[a-zA-Z0-9-]+}"
+	witStr := "{witid:[a-zA-Z0-9-.]+}"
```
AlCutter, line 118: This looks like it'll allow any character one or more times, is that intended?

distributor/cmd/main.go
```diff
@@ -0,0 +1,140 @@
...
+func main() {
+	flag.Parse()
+	ctx := context.Background()
+
+	// This error group will be used to run all top level processes.
+	// If any process dies, then all of them will be stopped via context cancellation.
+	g, ctx := errgroup.WithContext(ctx)
+	httpListener, err := net.Listen("tcp", *addr)
+	if err != nil {
+		glog.Exitf("failed to listen on %q", *addr)
```
AlCutter, line 63: nit: Capital (since it's not an error)

distributor/api/http.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Package api provides the API endpoints for the distributor.
+package api
+
+const (
+	// HTTPGetCheckpointN is the path of the URL to get a checkpoint with
+	// at least N signatures.  The placeholders are:
+	//  * first position is for the logID (an alphanumeric string)
+	//  * second position is the number of signatures required
+	HTTPGetCheckpointN = "/distributor/v0/logs/%s/checkpoint.%s"
```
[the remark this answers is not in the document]
mhutchinson: Nah, because what this gets subbed for is `{numsigs:\\d+}` which ain't a number.

distributor/cmd/internal/http/server.go
```diff
@@ -0,0 +1,136 @@
...
+		http.Error(w, fmt.Sprintf("failed to convert log list to JSON: %v", err), http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Type", "text/json")
+	w.Write(logList)
+}
+
+// RegisterHandlers registers HTTP handlers for witness endpoints.
+func (s *Server) RegisterHandlers(r *mux.Router) {
+	logStr := "{logid:[a-zA-Z0-9-]+}"
```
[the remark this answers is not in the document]
mhutchinson: Sure, but this block is only invoked once so I'm nay too worried about it. What's your concern? I feel like having the declaration of these strings being right next to where they're used (once) is better than having them further away from each other.

distributor/cmd/internal/http/server.go
mhutchinson: Good call, I've escaped the dot for now. This is a bit hacky really. Do we have a better dictionary for what might be legal characters in a vkey name?

distributor/cmd/main.go
mhutchinson: Done.

distributor/cmd/internal/http/server.go
mhutchinson: https://cs.opensource.google/go/x/mod/+/refs/tags/v0.10.0:sumdb/note/note.go;drc=7c05a442b7c1d1a107879b4a090bb5a38d3774a1;l=238

mhutchinson:
@AlCutter PTAL. I've left some of the comment threads open where I haven't made changes based on your comment.

distributor/cmd/internal/http/server.go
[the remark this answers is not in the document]
AlCutter: No real concern, just pointing out that they're (at least logically) consts is all :)

AlCutter (approved):

PR #788, opened 2023-04-13
Fix for Dockerfile smell DL3006

grosa1:
Hi!
The Dockerfile placed at "feeder/cmd/pixel_bt_feeder/Dockerfile" contains the best practice violation [DL3006](https://github.com/hadolint/hadolint/wiki/DL3006) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3006 occurs when the image tag for the base image is missing.
In this pull request, we propose a fix for that smell generated by our fixing tool. We have verified that the patch is correct before opening the pull request.
To fix this smell, specifically, we use a heuristic approach that selects the most probable version tag for the given base image. In detail, it selects the most recent image tag which corresponds to the same image digest that currently corresponds to the "latest" tag, used by default when pulling the image from DockerHub if a specific image tag is missing.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### DIFF
--- a/feeder/cmd/pixel_bt_feeder/Dockerfile
+++ b/feeder/cmd/pixel_bt_feeder/Dockerfile
@@ -19,7 +19,7 @@ COPY . .
 RUN go build -o /build/bin/pixel_bt_feeder ./feeder/cmd/pixel_bt_feeder
 
 # Build release image
-FROM alpine
+FROM alpine:3.17.3
 
 COPY --from=builder /build/bin/pixel_bt_feeder /bin/pixel_bt_feeder
 ENTRYPOINT ["/bin/pixel_bt_feeder"]


mhutchinson:
Thanks for the PR. I understand the intent of this lint rule, but I need a little time to think over what the correct solution is. I'm currently ambivalent. On one hand tracking the latest version of a base linux image seems quite reasonable; it's unlikely (though possible) that there will be any changes in that base image that would break the functionality we use of it. On the other hand, I don't like that this is not at all hermetic, thus an image built using the same version of our git repo could be very different depending on _when_ the build was performed (or more precisely/worryingly, whatever docker hub returns at that point, which now I think about it could be an attack vector in another way).

Looking at the [implementation of DL3006](https://github.com/hadolint/hadolint/blob/master/src/Hadolint/Rule/DL3006.hs#L20) it seems like we could pass this check by any of the following labels:
 - `FROM alpine:3.17.3`
 - `FROM alpine:3.17`
 - `FROM alpine:3`
 - `FROM alpine:latest`

The last one is equivalent to what we currently have, though the linter would pass (if I'm reading the rule correctly). I'm trying to weigh up whether this is still a Good Thing because at least the choice to track "latest" was explicitly written down. I guess it depends on whether the rule is trying to catch "unstable references are bad", or "developers should be explicit about how unstable they want their references to be". @grosa1, any thoughts?

Note that any choices other than the first one (and actually, including the first one if labels/tags are mutable) still violate the hermetic build ideal.

grosa1:
Hi @mhutchinson,
thanks for the response. 
I think that your point is correct. In my opinion, all the options excluding the tag `latest` could be useful in terms of keeping the build reliable as at least the major version of alpine is kept. For example, with `alpine:3` the minor updates and patches are not fixed to a specific version.
In addition, using the tag `latest`, leads to another smell detected by hadolint, [DL3007](https://github.com/hadolint/hadolint/wiki/DL3007). In that case, it could be that the developer can choose the tag `latest` to explicitly tell that the base image has to get always the latest version available.
The same thing can be in your case, if there is no explicit tag to get always the latest version, it is ok then. The detected smell, in that case, is a false positive

mhutchinson:
@grosa1 thanks for following up. I'll close this PR as #789 fixes this across the repo and also instructs dependabot to update these image tags. This gives us the benefit of having explicit versions in the repo, but avoids the problem that these explicit versions can get wildly out of date and miss security patches.

PR #797, opened 2023-04-19
Deleted a huge amount of witness and feeder code

mhutchinson:
This has moved to transparency-dev/witness


AlCutter (approved):
Does the top-level README.md need updating to remove/update the bit about the `/witness` directory?

Also, is it worth dropping a small `witness/README.md` file with a small `GOTO github.com/transparency-dev/witness` pointer?

AlCutter:
> I've updated the top-level README with a pointer, but I'm not in love with the ~empty directory pointing to another repo. I understand the intent, but I worry that a lot of the pointers you're trying to redirect may point further into the paths anyway (e.g. witness/golang/)

I can live with that, ta.

AlCutter (approved):

roger2hk:
The top level cloud build yaml file was removed in this PR. It breaks the GCB on master branch.

https://github.com/google/trillian-examples/commits/master

mhutchinson:
> The top level cloud build yaml file was removed in this PR. It breaks the GCB on master branch.
> 
> https://github.com/google/trillian-examples/commits/master

I've disabled the trigger that was trying to build using the docker images using the rules in `cloudbuild_docker.yaml`. @roger2hk, does this address your comment, or are there other errors?

roger2hk:
Yes if it's disabled in the master branch, so we won't see the red signal again.

PR #813, opened 2023-05-23
docs: fix docker-compose container name mismatch

aquelegustavo:
The docker-compose file uses "-" in the container name, adjusting the docs to make it easier for a anyone to run the tool.

Thans guys, 
You are amazing <3

### DIFF
--- a/helloworld/README.md
+++ b/helloworld/README.md
@@ -36,7 +36,7 @@ export TREE_ID=$(go run github.com/google/trillian/cmd/createtree --admin_server
 ```bash
 docker compose up -d
 # Wait a second for MySQL to start up between these commands
-curl https://raw.githubusercontent.com/google/trillian/master/storage/mysql/schema/storage.sql | docker exec -i helloworld_db_1 mysql -pzaphod -Dtest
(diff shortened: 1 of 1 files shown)


mhutchinson (approved):
Thanks for this. The instructions should be for a recent version of `docker compose` so your changes make sense. If line 37 had been `docker-compose` then the underscore naming convention would have been reasonable.

mhutchinson:
@roger2hk do you still want changes made to this PR? It's been a little over a month since I wrote it, but I think the underscores are a result of me testing this on a machine that only had `docker-compose` available and not fixing all of the instructions afterwards.

roger2hk (approved):

roger2hk:
> @roger2hk do you still want changes made to this PR? It's been a little over a month since I wrote it, but I think the underscores are a result of me testing this on a machine that only had `docker-compose` available and not fixing all of the instructions afterwards.

It's all good if Compose V2 is expected.

```
Compose V1 no longer receives updates and will not be available in new releases of Docker Desktop after June 2023.

Compose V2 is included with all currently supported versions of Docker Desktop. For more information, see [Migrate to Compose V2](https://docs.docker.com/compose/migrate).
```

https://docs.docker.com/compose/reference/

PR #834, opened 2023-06-22
Fix health check for mariadb instances

mhutchinson:
The vanilla image does not bundle a mysql command, and thus the
healthcheck was failing. The rpi version bundles both mysql and mariadb.
Using mariadb as the healthcheck command works on all platforms.

Also pinned the mariadb version as per best practice and to prevent
non-hermetic future breakages.


roger2hk (approved):
You may want to replace `MYSQL_*` with `MARIADB_*` equivalent variables.

mhutchinson:
> You may want to replace `MYSQL_*` with `MARIADB_*` equivalent variables.

Done!

roger2hk:
Apologies. The change was not applied to the following file.

https://github.com/google/trillian-examples/blob/b2c10cf095a9aaa9521716a3502f8cdc1bf7764e/distributor/docker-compose.yaml#L8-L11

mhutchinson:
I've sent #835 to fix the missing ones.

PR #836, opened 2023-06-28
Added test for getting checkpoint that has been available

mhutchinson:
This is a missing feature at the moment; the current behaviour of
getCheckpointN looks at the latest checkpoints for each witness and
merges them. This doesn't catch the case where witnesses are not aligned
at the point where the getCheckpointN is called.

### DIFF
--- a/distributor/cmd/internal/distributor/distributor_test.go
+++ b/distributor/cmd/internal/distributor/distributor_test.go
@@ -667,6 +667,137 @@ func TestGetCheckpointN(t *testing.T) {
 	}
 }
 
+func TestGetCheckpointNHistoric(t *testing.T) {
+	ws := map[string]note.Verifier{
+		"Whittle": witWhittle.verifier,
+		"Wattle":  witWattle.verifier,
+		"Waffle":  witWaffle.verifier,
+	}
+	ls := map[string]distributor.LogInfo{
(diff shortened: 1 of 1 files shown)


distributor/cmd/internal/distributor/distributor_test.go
```diff
@@ -667,6 +667,137 @@ func TestGetCheckpointN(t *testing.T) {
 	}
 }
 
+func TestGetCheckpointNHistoric(t *testing.T) {
+	ws := map[string]note.Verifier{
+		"Whittle": witWhittle.verifier,
```
AlCutter, line 672: May I respectfully request 3 witness/variable names that are quite a bit more visually distinct? :)

Somewhat embarrassingly, it took me a moment to notice in the test data structs that there were 3 witnesses, rather than 2 with one sending an updated checkpoint :)

AlCutter (approved):
Good call, love the TODO calling out how it _should_ be.

Just one request from my old eyes here 👀 to make it easier to grok what's happening.

distributor/cmd/internal/distributor/distributor_test.go
mhutchinson: Should have gone to SpecSavers :-)

Changed to animal names.

PR #852, opened 2023-07-24
Added tool for creating tile-based logs from cloned log

mhutchinson:
This tool allows a file-based mirror of any cloneable log to be created. This required a change to Integrate to remove the requirement for a prior checkpoint. This has been replaced with the fromSize. In the case of mirrors, we can't create and sign a valid checkpooint of size 0, which is what the integrate command had assumed was possible from its previous usage.

TODO: write up instructions for setting up a filesystem that can handle this without running out of space / inodes.

### DIFF
--- a/serverless/cmd/clone2serverless/main.go
+++ b/serverless/cmd/clone2serverless/main.go
@@ -118,9 +118,13 @@ func main() {
 	rCtxCancel()
 
 	glog.V(1).Infof("Integrating leaves [%d, %d)", fromSize, toSize)
-	log.Integrate(ctx, fromSize, logRoot, rfc6962.DefaultHasher)
+	if _, err := log.Integrate(ctx, fromSize, logRoot, rfc6962.DefaultHasher); err != nil {
+		glog.Exitf("Failed to integrate leaves: %v", err)
+	}
 	glog.V(1).Info("Writing checkpoint")
-	logRoot.WriteCheckpoint(ctx, inputCheckpointRaw)
+	if err := logRoot.WriteCheckpoint(ctx, inputCheckpointRaw); err != nil {
+		glog.Exitf("Failed to write checkpoint: %v", err)
+	}
 	glog.V(1).Infof("Successfully created tlog with %d leaves at %q", toSize, *outputRoot)
 }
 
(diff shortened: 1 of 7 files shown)


serverless/cmd/clone2serverless/main.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	glog.V(1).Infof("Successfully created tlog with %d leaves at %q", toSize, *outputRoot)
+}
+
+// initStorage loads or creates the tile-based log storage in the given output directory
+// and returns the storage, along with the tree size currently persisted.
+func initStorage(logVerifier note.Verifier, outputRoot string) (*fs.Storage, uint64, error) {
+	var logRoot *fs.Storage
+	var fromSize uint64
+	if fstat, err := os.Stat(outputRoot); os.IsNotExist(err) {
```
AlCutter, line 136: I think this misses the case where `err != nil` but does not represent a non-existent file - it _might_ not matter since we'll likely just error out lower down when we try to open the log files, but maybe explicitly logging & bailing might be better?

serverless/experimental/wasm/main.go
```diff
@@ -126,7 +126,7 @@ func integrate() js.Func {
 		}
 
 		ctx := context.Background()
-		newCp, err := log.Integrate(ctx, *cp, logStorage, rfc6962.DefaultHasher)
+		newCp, err := log.Integrate(ctx, *&cp.Size, logStorage, rfc6962.DefaultHasher)
```
AlCutter, line 129: Does just `cp.Size` not work?

AlCutter (approved):

serverless/experimental/wasm/main.go
mhutchinson: Hah, good spot. I think that must have been VS code suggesting an auto complete and I really didn't notice it adding the ampersand. It must have assumed the pointer dereference was super important.

serverless/cmd/clone2serverless/main.go
mhutchinson: Good call, added this check in.

AlCutter (approved):

PR #862, opened 2023-08-10
Create scorecard.yml

mhutchinson:
Following the instructions for [scorecard installation](https://github.com/ossf/scorecard-action#installation).


roger2hk (approved):
Although the actions are not using the latest version, I'll let Dependabot to catch them up.

PR #977, opened 2023-10-16
Support clone tool running continuously

mhutchinson:
This improves the health appearance for this tool when running happily inside docker. Before this change, it exits after cloning, and is then restarted by docker. This makes it hard at a glance to distinguish crash looping from normal operation.

### DIFF
--- a/clone/cmd/sumdbclone/sumdbclone.go
+++ b/clone/cmd/sumdbclone/sumdbclone.go
@@ -72,11 +72,10 @@ func main() {
 	for {
 		select {
 		case <-ticker.C:
-			// Wait until the next tick.
+			cloneAndVerify(ctx, client, db)
 		case <-ctx.Done():
 			glog.Exit(ctx.Err())
 		}
-		cloneAndVerify(ctx, client, db)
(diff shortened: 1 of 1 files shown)


clone/cmd/sumdbclone/sumdbclone.go
```diff
@@ -63,6 +64,24 @@ func main() {
...
 		Timeout: *timeout,
 	})
+	cloneAndVerify(ctx, client, db)
+	if *pollInterval == 0 {
+		return
+	}
+	ticker := time.NewTicker(*pollInterval)
+	for {
+		select {
+		case <-ticker.C:
```
phbnf, line 74: Why doesn't it call cloneAndVerify directly here?

phbnf (approved):

clone/cmd/sumdbclone/sumdbclone.go
mhutchinson: No good reason other than consistency with some other code I'd previously written. Agree that this way is easier to read. Fixed.

PR #983, opened 2023-10-23
Removed obsolete sumdbaudit command

mhutchinson:
This has been replaced with sumdbverify, which now has useful information that was in the sumdbaudit README incorporated


mhutchinson:
I hadn't realized that these tools had some entanglement. Switching to draft until I've bottomed this out.

clone/cmd/sumdbclone/internal/client/sumdb.go
```diff
@@ -0,0 +1,164 @@
...
+func (c *SumDBClient) tilePath(offset int) string {
+	nStr := fmt.Sprintf("%03d", offset%pathBase)
+	for offset >= pathBase {
+		offset /= pathBase
+		nStr = fmt.Sprintf("x%03d/%s", offset%pathBase, nStr)
+	}
+	return nStr
+}
+
+func dataToLeaves(data []byte) [][]byte {
```
AlCutter, line 123: Is this the same as `bytes.Split(data, []byte{'\n'})`?

clone/cmd/sumdbclone/internal/client/sumdb.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			glog.Errorf("resp.Body.Close(): %v", err)
+		}
+	}()
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("GET %v: %v", target, resp.Status)
+	}
+	data, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
```
AlCutter, line 159: could this just directly `return io.ReadAll(...)` since you're not decorating the error?

AlCutter (approved):
Just a couple of little suggestions in what (now) looks like pre-existing code - the link I clicked on to review this brought me to just the 2nd commit, and the entirety of those re-added files appeared to be new.

clone/cmd/sumdbclone/internal/client/sumdb.go
mhutchinson: Done.

clone/cmd/sumdbclone/internal/client/sumdb.go
mhutchinson: No, but I've replaced it with something similar. Annoying to split on "\n\n" and then need to re-add it, but I'll survive.

PR #1045, opened 2023-12-27
Adapt ctclone for 'coerced get-entries' in Let's Encrypt and other CT Logs

### LINKED ISSUE #1044: ctclone Fails with Incomplete CT Log Responses
### Issue Description

The ctclone tool encounters failures when Certificate Transparency (CT) logs return fewer entries than requested. This issue becomes problematic during periodic ctclone operations or at the ends of the logs. At certain points, the position in the cloned log may align in such a way that the CT log returns less than the expected number of entries.

### Specific Example

An instance of this issue can be observed with Sectigo's Sabre 2025h2. When a "negotiated" page size of 256 is used, the log [returns only 165 entries for a specific range](https://sabre2025h2.ct.sectigo.com/ct/v1/get-entries?start=52827&end=53082). This response unexpectedly brings the position to 52992 (207 * 256), causing ctclone to fail with the following error:

```
worker 0: Retryable error getting data: "LeafFetcher.Batch(52827, 256): wanted 256 leaves but got 165"
```

### Additional Context

Further details about this behavior in CT logs can be found in the discussion on [Let's Encrypt's forum regarding the 'coerced get-entries' feature](https://community.letsencrypt.org/t/enabling-coerced-get-entries/114436).

### Important Note

The proposed solution covers only the cases where the actual log's page size is a multiple of the requested batch size. It is assumed that the negotiated batch size requested by ctclone aligns with this multiple. If the actual page size does not conform to this pattern, additional adjustments may be required.

Edited: formatting.

tired-engineer:
This commit updates ctclone to handle the 'coerced get-entries' feature introduced in Let's Encrypt's Certificate Transparency logs. Previously, ctclone did not account for the log's behavior of truncating responses to align with cacheable boundaries.

Fixes #1044

### DIFF
--- a/clone/cmd/ctclone/ctclone.go
+++ b/clone/cmd/ctclone/ctclone.go
@@ -106,7 +106,7 @@ type ctFetcher struct {
 }
 
 // Batch provides a mechanism to fetch a range of leaves.
-// Enough leaves are fetched to fully fill `leaves`, or an error is returned.
+// Enough leaves are fetched to fully fill `leaves`, or an error is returned, unless the page is coerced.
 // This implements batch.BatchFetch.
 func (cf ctFetcher) Batch(start uint64, leaves [][]byte) error {
 	// CT API gets [start, end] not [start, end).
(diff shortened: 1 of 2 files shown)


tired-engineer:
Found the subsequent issue in batching process, closing for now.

PR #1055, opened 2024-01-16
Adapt ctclone for 'coerced get-entries' in Let's Encrypt and other CT Logs

tired-engineer:
Adapt ctclone to handle the 'coerced get-entries' feature by: trying to retrieve just one request, checking the number of returned results, either continuing with the rest workers, or stopping fetching, storing the incomplete tile, exiting delegating orchestration system to start the process again.

Fixes https://github.com/google/trillian-examples/issues/1044

### DIFF
--- a/clone/cmd/ctclone/ctclone.go
+++ b/clone/cmd/ctclone/ctclone.go
@@ -91,24 +91,24 @@ type ctFetcher struct {
 // Batch provides a mechanism to fetch a range of leaves.
 // Enough leaves are fetched to fully fill `leaves`, or an error is returned.
 // This implements batch.BatchFetch.
-func (cf ctFetcher) Batch(start uint64, leaves [][]byte) error {
+func (cf ctFetcher) Batch(start uint64, leaves [][]byte) (uint64, error) {
 	// CT API gets [start, end] not [start, end).
 	last := start + uint64(len(leaves)) - 1
 	data, err := cf.f.GetData(fmt.Sprintf("ct/v1/get-entries?start=%d&end=%d", start, last))
 	if err != nil {
-		return fmt.Errorf("fetcher.GetData: %w", err)
+		return 0, fmt.Errorf("fetcher.GetData: %w", err)
 	}
 	var r getEntriesResponse
 	if err := json.Unmarshal(data, &r); err != nil {
-		return fmt.Errorf("json.Unmarshal of %d bytes: %w", len(data), err)
+		return 0, fmt.Errorf("json.Unmarshal of %d bytes: %w", len(data), err)
 	}
 	if got, want := len(r.Leaves), len(leaves); got != want {
-		return fmt.Errorf("wanted %d leaves but got %d", want, got)
+		return uint64(len(r.Leaves)), fmt.Errorf("wanted %d leaves but got %d", want, got)
 	}
 	for i, l := range r.Leaves {
 		leaves[i] = l.Data
 	}
-	return nil
+	return uint64(len(r.Leaves)), nil
 }
 
 func (cf ctFetcher) latestCheckpoint() (CTCheckpointResponse, error) {
--- a/clone/internal/download/batch.go
+++ b/clone/internal/download/batch.go
@@ -17,6 +17,7 @@ package download
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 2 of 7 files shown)


clone/cmd/ctclone/ctclone.go
```diff
@@ -108,24 +108,24 @@ type ctFetcher struct {
 // Batch provides a mechanism to fetch a range of leaves.
 // Enough leaves are fetched to fully fill `leaves`, or an error is returned.
 // This implements batch.BatchFetch.
-func (cf ctFetcher) Batch(start uint64, leaves [][]byte) error {
+func (cf ctFetcher) Batch(start uint64, leaves [][]byte) (uint64, error) {
```
AlCutter, line 111: @mhutchinson what's the reason the `Batch` function signature takes an "OUT" parameter rather than being something like `f(start, N uint64) ([][]byte, error) and just returning what was available`?

It feels like the `leaves` param is kinda doing a slightly unusual "double duty" here: informing the implementation of the max to fetch as well as being the container to put them in, which then leads to having to special case "short" reads.

clone/internal/download/batch.go
```diff
@@ -78,29 +86,42 @@ func Bulk(ctx context.Context, first, treeSize uint64, batchFetch BatchFetch, wo
 			return
 		case r = <-rangeChans[i]:
 		}
-		if r.err != nil {
-			rc <- BulkResult{
-				Leaf: nil,
-				Err:  r.err,
+
+		if i == 0 && !alignmentDone && r.err == nil {
```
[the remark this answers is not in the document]
mhutchinson: Yeah this is a known brittleness, but a pragmatic step forward to fix the immediate issue. See discussion in #1044 about how this running in something like docker would restart the job and hopefully lead to snapping into alignment. The current code bakes in the batch size quite early on, and this is only intended to fix a specific observed issue and not all conceivable alignment issues.

If we want this to be fully general in the face of logs that regularly don't respect request parameters then we need to take a more comprehensive step back and rewrite.

clone/cmd/ctclone/ctclone.go
mhutchinson: I can't remember if this was a performance thing, or trying to make this look something like `io.Reader`. I don't think the original contract was necessarily bad, nor is this change (which actually brings it closer to `io.Reader`, nor your suggestion.

clone/internal/download/batch.go
```diff
@@ -48,25 +47,34 @@ func Bulk(ctx context.Context, first, treeSize uint64, batchFetch BatchFetch, wo
 	rangeChans := make([]chan workerResult, workers)
 
 	increment := workers * batchSize
+
+	waitCh := make(chan struct{})
```
mhutchinson, line 51: Instead of creating this channel and then doing alignment inside the concurrency below, I imagined that this change would do the alignment in the main thread here, and once it had got alignment, would then kick off the concurrency by continuing on with an updated `start`. Is this possible?

I have to jump to meetings now, but LMK if this doesn't make sense.

mhutchinson:
Hi @tired-engineer, I've finally emerged from the deep rabbit hole of "other stuff" that I was context loaded on, and have given this a bit of time. Apologies for the delay. What do you think about https://github.com/google/trillian-examples/commit/6f6ea6083d0da672d382331304dd1a0d7def3806 ?

This unwinds the logic for making the first goroutine special (and the other goroutines knowing it). Instead, the main thread does the work. After doing this change, it was only a small amount more code to make it so that it self heals, so it actually tries the alignment twice if the first alignment doesn't get a full batch.

tired-engineer:
Sorry @mhutchinson, for such a long delay. I'd say it's good (at least for me).

mhutchinson:
> Sorry @mhutchinson, for such a long delay. I'd say it's good (at least for me).

No worries, thanks for coming back! If you want to take credit for your work on this then please:
1. cherry pick my changes into this PR
2. rebase
3. Mark the PR as ready for review

We'll get this merged :-)

If I've not heard from you by the end of the month then I'll do this myself and we'll give you credit via the commit message and changelog.

mhutchinson (approved):

mhutchinson:
@tired-engineer thanks for your work on this!

PR #1189, opened 2024-07-16
Fix extra spaces on last leaf in bundle

### LINKED ISSUE #1188: "failed to clone and verify log" error while using `sumdbclone` example
Hey, first of all I appreciate this is example code and I shouldn't expect too much, so thank you for the examples provided here.

Every time I've tried to clone the sumdb log I've hit this same sort of error, I've tried to empty out the database tables and start fresh and eventually hit the same error. The initial clone seems to work fine and the `leaves` table seems to fill up ~7.5GB of content, but after the initial clone the verify stage always fails with errors like the below:

`sumdbclone.go:88] Failed to clone log: failed to clone and verify log: computed root 7e3499466c71a429de6508bb87514f506e3e7a10d47fc2b172cc610aa2dd8cc0 != provided checkpoint aef05edcb9f3f3a7198f6e5e891f97c535d3ad2aa56ee4247614b1b5850bdd1d for tree size 28371487`

Any pointers where I'm going wrong are appreciated 🙏🏻

mhutchinson:
Fixes #1188.

### DIFF
--- a/clone/cmd/sumdbclone/internal/client/sumdb.go
+++ b/clone/cmd/sumdbclone/internal/client/sumdb.go
@@ -121,10 +121,14 @@ func (c *SumDBClient) tilePath(offset int) string {
 	return nStr
 }
 
+// dataToLeaves splits the raw bytes received when requesting a data
+// tile into the constituent leaf data chunks. The splitting is performed
+// based on two consecutive newlines, and each returned leaf terminates
+// with a single newline.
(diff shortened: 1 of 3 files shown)


AlCutter (approved):

mhutchinson:
I've got about 15mins left of running this locally to confirm it fixes the clone tool, and then I'll submit.

PR #1232, opened 2024-09-23
Clone tool: use max size uint in DB schema

### LINKED ISSUE #1229: Signed id/size in database tables
Google's Argon and Xenon logs contain around 2.3B records.  A signed integer can handle the range +/-2.1B.

When cloning Argon I get this message:
```
E0921 06:55:56.360252  685659 database.go:135] tx.Exec(): Error 1264 (22003): Out of range value for column 'id' at row 1
E0921 06:55:56.360316  685659 database.go:135] tx.Exec(): Error 1264 (22003): Out of range value for column 'id' at row 1
E0921 06:55:56.360383  685659 database.go:135] tx.Exec(): Error 1264 (22003): Out of range value for column 'id' at row 1
...
```

By inspection it looks like all that is needed is to alter the leaves and checkpoints tables to use integer unsigned for `id` and `size` respectively.  The code all seems to use `uint64` so once the database has an unsigned int we should be ok up to 4.2B entries.  I'm backing up my database before altering the table, so have not tested that change yet.

Here is where the keys are defined as signed:
https://github.com/google/trillian-examples/blob/master/clone/logdb/database.go#L59-L62

mhutchinson:
This fixes #1229.

### DIFF
--- a/clone/logdb/database.go
+++ b/clone/logdb/database.go
@@ -56,10 +56,10 @@ func NewDatabaseDirect(db *sql.DB) (*Database, error) {
 }
 
 func (d *Database) init() error {
-	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS leaves (id INTEGER PRIMARY KEY, data BLOB)"); err != nil {
+	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS leaves (id BIGINT UNSIGNED PRIMARY KEY, data BLOB)"); err != nil {
 		return err
 	}
-	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS checkpoints (size INTEGER PRIMARY KEY, data BLOB, compactRange BLOB)"); err != nil {
+	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS checkpoints (size BIGINT UNSIGNED PRIMARY KEY, data BLOB, compactRange BLOB)"); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


clone/logdb/database.go
```diff
@@ -56,10 +56,10 @@ func NewDatabaseDirect(db *sql.DB) (*Database, error) {
 }
 
 func (d *Database) init() error {
-	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS leaves (id INTEGER PRIMARY KEY, data BLOB)"); err != nil {
+	if _, err := d.db.Exec("CREATE TABLE IF NOT EXISTS leaves (id BIGINT UNSIGNED PRIMARY KEY, data BLOB)"); err != nil {
```
AlCutter, line 59: Should we try `ALTER TABLE` if the tables do exist (or, perhaps better, print a message with the commands?)

AlCutter (approved):

clone/logdb/database.go
mhutchinson: Without having a way to query the schema to see what the current column type is defined to be, this would turn into log spam even for users that have already run the commands. Adding code the query the schema is a bit much and is likely to make this more brittle and less portable between MySQL and sqlite (https://stackoverflow.com/questions/1215368/how-to-get-the-mysql-table-columns-data-type).

If we're going to do schema migration properly, we should do it properly and use github.com/golang-migrate/migrate.

PR #1254, opened 2024-10-21
Fix sumdb batchmap example

mhutchinson:
This was previously based on top of the deleted sumdbaudit tool. The
sumdbaudit tool used to have its own DB format, but was obsoleted by the
multi-purpose clone tool and the sumdbverify command. When the audit
command was deleted, apparently this dusty experimental corner got
forgotten about, and so this map could no longer be built because the
data source it required for a log clone was no longer available.

This change updates the sumdb map in the minimal way to make it usable
from the sumdb data cloned using the clone tool. This requires an extra
step to parse the generic row format for leaves and turn this into the
expanded data that the map needs.

Further work is likely here to make the map more useful and easier to
understand. Documentation, refactoring and revisiting the need for the
salt function are currently top of mind.


mhutchinson:
Weird, it ran locally. I'll take another look tomorrow.

mhutchinson:
It was the tests that were broken. I've probably overdone it changing the main pipeline to fix them, but it reads a bit better this way, so YOLO.

AlCutter (approved):

PR #1267, opened 2024-11-12
[Dependabot] Use multi-directory to reduce duplicate config

mhutchinson:
This will also make it so that dependabot will update all of these grouped dependencies in a single PR. See https://github.com/google/trillian/pull/3672 for the first application of this in the trillian repo. This will massively cut down on maintenance costs and make PR history much easier to digest.


AlCutter (approved):
❤️

PR #1330, opened 2025-06-18
Use vindex from transparency-dev

mhutchinson:
This deletes the local version of the library and uses the transparency-dev clone of it. The deps and go version were bumped as part of this and seem to be part of a transitive dependency chain.


AlCutter (approved):
Is the longer-term plan to move all the vindex example commands over to `incubator` too?

